PR #717, opened 2018-01-18
Work around incorrect pattern match parsing

mikusp:
Incorrect pattern matches, like `a + b = "foo"` will now show a parse error


wdanilo:
cool!

Issue #577, opened 2017-12-18
Preserve last opened function when working with multiple files


kustosz:
Currently, when I have two luna files and switch between them, I get moved to the main function, regardless of where I was before the switch. It's a major distraction.

PR #735, opened 2018-01-19
Retrieve last visited location when file changes. #577

ljkania:
We should think about what should happen when file is opened for the first time in the session: should we open last location or `main` function? Also @kustosz mentioned something about not sharing this information with project - it is saved in project for now, if we need to separate it then we need to talk about this. Anyway: shouldn't whole this file not be shared? I mean when you develop you do not want to share your every move on graph to the github - so I guess this whole file should simply be included in `.gitignore`. @kustosz @wdanilo ?

### DIFF
--- a/libs/batch/plugins/luna-empire/src/Empire/Server/Graph.hs
+++ b/libs/batch/plugins/luna-empire/src/Empire/Server/Graph.hs
@@ -23,7 +23,7 @@ import           Data.ByteString.Lazy                    (fromStrict)
 import           Data.Char                               (isUpper)
 import qualified Data.HashMap.Strict                     as HashMap
 import qualified Data.IntMap                             as IntMap
-import           Data.List                               (break, find, partition, sortBy, (++))
+import           Data.List                               (break, find, partition, sortBy)
 import           Data.List.Split                         (splitOneOf)
 import qualified Data.Map                                as Map
 import           Data.Maybe                              (isJust, isNothing, listToMaybe, maybeToList)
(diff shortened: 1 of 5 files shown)


libs/batch/plugins/luna-empire/src/Empire/Server/Graph.hs
```diff
@@ -23,7 +23,7 @@ import           Data.ByteString.Lazy                    (fromStrict)
 import           Data.Char                               (isUpper)
 import qualified Data.HashMap.Strict                     as HashMap
 import qualified Data.IntMap                             as IntMap
-import           Data.List                               (break, find, partition, sortBy)
+import           Data.List                               (break, find, partition, sortBy, (++))
```
wdanilo: You should bnever use `++`, please use `<>` instead.

libs/batch/plugins/luna-empire/src/Empire/Server/Graph.hs
```diff
@@ -172,40 +172,50 @@ getProjectPathAndRelativeModulePath modulePath = do
...
-getMainLocation :: GraphLocation -> Empire (Maybe GraphLocation)
-getMainLocation gl = do
+saveSettings :: GraphLocation -> LocationSettings -> GraphLocation -> Empire ()
+saveSettings gl settings newGl = handle (\(e :: SomeException) -> return ()) $ do
+    bc    <- Breadcrumb.toNames <$> Graph.decodeLocation gl
+    newBc <- Breadcrumb.toNames <$> Graph.decodeLocation newGl
+    let filePath        = gl    ^. GraphLocation.filePath
+        newFilePath     = newGl ^. GraphLocation.filePath
+        lastBcInOldFile = if filePath == newFilePath then newBc else bc
+    liftIO $ do
```
wdanilo: We should almost never use `liftIO` in our high level code. Instead the functions `Project.updateLocationSettings` etc should already be lifted.

libs/batch/plugins/luna-empire/src/Empire/Server/Graph.hs
```diff
@@ -172,40 +172,50 @@ getProjectPathAndRelativeModulePath modulePath = do
...
+getClosestBcLocation gl (Breadcrumb (nodeName:newBcItems)) = do
     g <- Graph.getGraph gl
-    return $ maybe def (\nid -> Just $ gl & GraphLocation.breadcrumb . Breadcrumb.items %~ (<> [Breadcrumb.Definition nid])) $ findMainNodeId g
+    let mayN = find ((Just nodeName ==) . view Node.name) (g ^. GraphAPI.nodes)
+    case mayN of
+        Nothing -> return gl
+        Just n  -> do
+            let nid = n ^. Node.nodeId
+                bci = if n ^. Node.isDefinition then Breadcrumb.Definition nid else Breadcrumb.Lambda nid
+            getClosestBcLocation (gl & GraphLocation.breadcrumb . Breadcrumb.items %~ (++[bci])) $ Breadcrumb newBcItems
```
wdanilo: This is unreadable. I do not know what it does by lookign at it for 2 seconds. You should create some helper variables and name them nice to make the code easy to follow.

libs/batch/plugins/luna-empire/src/Empire/Server/Graph.hs
```diff
@@ -172,40 +172,50 @@ getProjectPathAndRelativeModulePath modulePath = do
...
+        when (filePath /= newFilePath) $ withJustM (getProjectPathAndRelativeModulePath newFilePath) $ \(cp, fp) ->
+            Project.updateCurrentBreadcrumbSettings cp fp newBc
+
+getClosestBcLocation :: GraphLocation -> Breadcrumb Text -> Empire GraphLocation
+getClosestBcLocation gl (Breadcrumb []) = return gl
+getClosestBcLocation gl (Breadcrumb (nodeName:newBcItems)) = do
     g <- Graph.getGraph gl
-    return $ maybe def (\nid -> Just $ gl & GraphLocation.breadcrumb . Breadcrumb.items %~ (<> [Breadcrumb.Definition nid])) $ findMainNodeId g
+    let mayN = find ((Just nodeName ==) . view Node.name) (g ^. GraphAPI.nodes)
+    case mayN of
```
wdanilo: You should be able to make it shorter by using function `maybe` here

libs/batch/plugins/luna-empire/src/Empire/Server/Graph.hs
```diff
@@ -172,40 +172,50 @@ getProjectPathAndRelativeModulePath modulePath = do
...
-    withJust mayProjectPathAndRelModulePath $ \(pf, mf) ->
-        liftIO $ Project.updateLocationSettings pf mf bc settings
-
-findMainNodeId :: Graph -> Maybe NodeId
-findMainNodeId g = view Node.nodeId <$> find ((Just "main" ==) . view Node.name) (g ^. GraphAPI.nodes)
-
-getMainLocation :: GraphLocation -> Empire (Maybe GraphLocation)
-getMainLocation gl = do
+saveSettings :: GraphLocation -> LocationSettings -> GraphLocation -> Empire ()
+saveSettings gl settings newGl = handle (\(e :: SomeException) -> return ()) $ do
```
mikusp: While we are at it, you could consider logging this exception somewhere

wdanilo:
@ljkania @kustosz lets talk about gitignoring the file live please!

PR #773, opened 2018-01-25
add '--version' for luna-studio

sylwiabr:


runner/src/StudioRunner.hs
```diff
@@ -147,6 +147,11 @@ version = do
     versionTxt <- versionText
     return $ fromText versionTxt
 
+printVersion :: (MonadRun m, MonadCatch m) => m ()
+printVersion = do
+    versionTxt <- catchAny versionText $ \e -> return "develop"
```
mikusp: Catch only isDoesNotExistError here

runner/src/StudioRunner.hs
```diff
@@ -147,6 +148,11 @@ version = do
     versionTxt <- versionText
     return $ fromText versionTxt
 
+printVersion :: (MonadRun m, MonadCatch m) => m ()
+printVersion = do
+    versionTxt <- catch versionText $ \e -> if (isDoesNotExistError e) then return "develop" else return . T.pack $ show e
```
piotrMocz: You can factor out the `return`:
```
return $ if (isDoesNotExistError e) then "develop" else T.pack (show e)
```

PR #779, opened 2018-01-29
Supervisor fixes

### LINKED ISSUE #701: Error: Another program is already listening on a port that one of our HTTP servers is configured to use
After running the `lunainstaller` binary, I tried to run the `luna-studio` binary inside my home directory, but it is throwing an error saying `Another program is already listening on a port that one of our HTTP servers is configured to use`

```
$ ~/.luna/bin/luna-studio/1.0/luna-studio 
unix:///tmp/supervisor.sock no such file
Error: Another program is already listening on a port that one of our HTTP servers is configured to use.  Shut this program down first before starting supervisord.
For help, use /tmp/.mount_luna-swK56od/usr/third-party/supervisord/supervisord -h
luna-studio: Received ExitFailure 2 when running
Shell command: /tmp/.mount_luna-swK56od/usr/third-party/supervisord/supervisord -n -c supervisord-package.conf
Run from: /tmp/.mount_luna-swK56od/usr/config/supervisor
```

I have checked my bound ports but none of them are listening to common ports like `3000` or `8000` or `8080` etc, nor any of the ports I've found specified in this repo - `8088`, `30530`, `30531`, `30532`.

I have checked that `supervisor` service is running:

```
$ pgrep -fl supervisor
27312 supervisord
```

There are no errors in `/var/log/supervisor/supervisord.log`

So which port should I clear up? And would it be possible to provide a more informative error message telling the end user exactly the port in question.

---

My system info:

```
$ uname -a
Linux 4.13.0-25-generic #29-Ubuntu SMP Mon Jan 8 21:14:41 UTC 2018 x86_64 x86_64 x86_64 GNU/Linux
```

mikusp:
This should fix luna/luna-manager#112, cause I suspect the cause is an unescaped space in ATOM_HOME. Maybe it will resolve #701, because supervisord no longer start its HTTP server on port 9101

### DIFF
--- a/supervisor/supervisord-package.conf
+++ b/supervisor/supervisord-package.conf
@@ -1,9 +1,6 @@
 [unix_http_server]
 file=/tmp/supervisor.sock   ; (the path to the socket file)
 
-[inet_http_server]          ; inet (TCP) server disabled by default
-port=127.0.0.1:9101         ; (ip_address:port specifier, *:port for all iface)
-
 [supervisord]
 logfile=%(ENV_LUNA_STUDIO_LOG_PATH)s/supervisord.log  ; (main log file;default $CWD/supervisord.log)
 logfile_maxbytes=50MB        ; (max main logfile bytes b4 rotation;default 50MB)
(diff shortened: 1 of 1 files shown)


supervisor/supervisord-package.conf
```diff
@@ -23,7 +20,7 @@ supervisor.rpcinterface_factory = supervisor.rpcinterface:make_main_rpcinterface
 serverurl=unix:///tmp/supervisor.sock ; use a unix:// URL  for a unix socket
 
 [program:luna-atom]
-environment=ATOM_HOME=%(ENV_LUNA_STUDIO_GUI_CONFIG_PATH)s,LD_LIBRARY_PATH=%(ENV_OLD_LIBPATH)s
+environment=ATOM_HOME="%(ENV_LUNA_STUDIO_GUI_CONFIG_PATH)s"
 command=%(ENV_LUNA_STUDIO_GUI_PATH)s -w %(ENV_LUNA_STUDIO_ATOM_ARG)s
```
piotrMocz: Should we escape this one as well?

Issue #737, opened 2018-01-19
Basics tutorial doesn't work on Windows 10


ricksanchez:
I just installed the Luna 1.0

I clicked in the BasicstTutorial
Click next, next, until I was requested to write circle and button "down" and enter
the circle node was not created and the "tab" menu is still displayed but it doesn't respond to esc/enter
the current text in the text bubble is "Click on the output port of the radius....", but at this point I cannot longer interact with the editor.
When I click next, a line from the node to the cursor appears and in the bubble "connecting the nodes, please wait...." and got stuck forever (until I click X and the tutorial finished).

crawfordcomeaux:
I have this same issue on Ubuntu 14.04. Should I open another issue for this?

wdanilo:
@crawfordcomeaux, no - it is related. However, please do provide your system info!

crawfordcomeaux:
Would you be willing to give me a set of commands to generate the info you'd like?

crawfordcomeaux:
My issue is different. The tutorial's really slow to respond, spending minutes at "please wait," typically at "Viewing the results" steps.

wdanilo:
crawfordcomeaux , sure, lets check it out! Please execute it when running Luna Studio and tutorials:
```less /proc/cpuinfo
free -mh
```
Moreover, what is your cpu usage when you observe the slowdown? Does it run faster if you wait for a while? Im asking it because Luna currently compiles stdlib in yhe background everytime you start it.

cguebert:
I have a similar issue, the circle node is never created, even if I click on the 'next' button. The tutorial is then in a dead end, and as all actions are blocked in the tutorial, I cannot progress.

ricardopieper:
Same problem: then I create the node, the tutorial asks me: Click on the output port of the radius.... but I can't do that.

Gurkenglas:
Here's a video reproducing this. https://streamable.com/eocf5

kustosz:
There is a workaround in the tutorial scenario for this. If you run the tutorial again (no need to update or anything, just select it again on the dashboard). Keeping this issue open for now, as the root is not solved yet, but you should all be able to use the tutorial.

PR #781, opened 2018-01-29
Fix tutorial on Windows. #737

ljkania:
Tutorial weren't working properly on windows because before node editor could make change restrictions were changed for the next step. I've redesigned this system and now we ensure that event will be handled and action will be performed via expected section in `guide.yml` file.

### DIFF
--- a/luna-studio/atom/lib/guide.coffee
+++ b/luna-studio/atom/lib/guide.coffee
@@ -143,27 +143,28 @@ module.exports =
                 nextStepNo = @nextStepNo
                 highlightedRect = hgElem.getBoundingClientRect()
                 if highlightedRect.width != 0 and highlightedRect.height != 0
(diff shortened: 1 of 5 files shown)


luna-studio/atom/lib/guide.coffee
```diff
@@ -143,27 +143,28 @@ module.exports =
                 nextStepNo = @nextStepNo
                 highlightedRect = hgElem.getBoundingClientRect()
                 if highlightedRect.width != 0 and highlightedRect.height != 0
+                    @buttonDoIt.show()
```
pmlodawski: if you will create a step without target (or target.action = 'proceed'), it will create two buttons: `buttonDoIt` and  `buttonContinue`

ljkania: ack.

PR #782, opened 2018-01-29
Flatten nested object tables

kustosz:
This improves the way our nested tables visualizations work. If you have a uniform list of objects like:
```
[ {foo: {foo1: ..., foo2: ...}, {bar: {bar1: ..., bar2: ...}}
, {foo: {foo1: ..., foo2: ...}, {bar: {bar1: ..., bar2: ...}}
, ...]
```
The way it used to work was you'd get two columns (`foo` and `bar`) and then each cell would become a tiny two-column table. This in turn resulted in bloating the table (the nested headers would appear for each table cell).

With this change, the structure gets visualized as a 4-column table, with `foo.foo1`, `foo.foo2`, `bar.bar1` and `bar.bar2` columns. No more redundancy in the structure!

### DIFF
--- a/luna-studio/atom/lib/visualizers/base/table/table.js
+++ b/luna-studio/atom/lib/visualizers/base/table/table.js
@@ -9,6 +9,30 @@
     return Object.keys(obj).length == keys.length && keys.every(k => obj.hasOwnProperty(k));
   };
 
+  var getAtNestedKey = function (data, key) {
+    var res = data;
+    key.forEach(function (k) { res = res[k]; });
+    return res;
+  }
+
+  var repNestedKey = function (key) {
+    return key.join(".");
+  }
+
+  var generateNestings = function (data, key) {
+    var first = getAtNestedKey(data[0], key);
+    if (!(first instanceof Object)) return [key];
+    var firstKeys = Object.keys(first);
+    var isNestable = data.every(obj => hasExactlyKeys(firstKeys, getAtNestedKey(obj, key)));
+    if (isNestable) {
+      var withNests = firstKeys.map(k => key.concat([k]));
+      var furtherNestings = withNests.map(k => generateNestings(data, k));
+      return [].concat.apply([], furtherNestings);
+    } else {
+      return [key];
+    }
+  }
+
   var isObjectMatrix = function (data) {
     var isList = Array.isArray(data) && data[0];
     if (!isList || !(typeof data[0] === "object"))  return false;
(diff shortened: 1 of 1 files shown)


luna-studio/atom/lib/visualizers/base/table/table.js
```diff
@@ -87,6 +115,7 @@
     } else if (data instanceof Object) {
       return "<td>" + genRowObjectTable(data, level + 1) + "</td>";
     } else {
+      if (data === undefined || data === null) data = "";
```
piotrMocz: Or: `var res = (data || "").toString();`

luna-studio/atom/lib/visualizers/base/table/table.js
```diff
@@ -9,6 +9,33 @@
...
+
+  var generateNestings = function (data, key) {
+    var first = getAtNestedKey(data[0], key);
+    if (!(first instanceof Object)) return [key];
+    var firstKeys = Object.keys(first);
+    var isNestable = data.every(obj => hasExactlyKeys(firstKeys, getAtNestedKey(obj, key)));
+    if (isNestable) {
+      var withNests = firstKeys.map(k => key.concat([k]));
+      var furtherNestings = withNests.map(k => generateNestings(data, k));
+      return [].concat.apply([], furtherNestings);
```
piotrMocz: I'm not sure I like this syntax very much. Confusing for future readers of the code, you have to explicitly know what it does. I wouldn't bet my head on it, but isn't it the same as:
```
return furtherNestings(reduce((x, y) => x.concat(y));
```
?
If yes, I would change it for the reader's convenience :)

luna-studio/atom/lib/visualizers/base/table/table.js
```diff
@@ -9,6 +9,33 @@
     return Object.keys(obj).length == keys.length && keys.every(k => obj.hasOwnProperty(k));
   };
 
+  var getAtNestedKey = function (data, key) {
+    var res = data;
+    key.forEach(function (k) { res = res[k]; });
```
piotrMocz: You can actually use the arrow syntax with statements, like: `key.forEach(k => { res = res[k]; });`

luna-studio/atom/lib/visualizers/base/table/table.js
kustosz: Not sure, isn't that slower somehow? Like multiple copies of the intermediate arrays being created?

luna-studio/atom/lib/visualizers/base/table/table.js
kustosz: Nope, I'd still like to have my `false`s and `0`s displayed like they are supposed to be.

luna-studio/atom/lib/visualizers/base/table/table.js
piotrMocz: I am not sure, I like to believe in reasonable compiler optimizations. You can let it be, but maybe write a comment what it does? I needed a few minutes to figure that out :D

luna-studio/atom/lib/visualizers/base/table/table.js
piotrMocz: Fair point, sir.

PR #792, opened 2018-02-02
Use new extensible snapshots

mikusp:
Motivation: https://www.fpcomplete.com/blog/2017/07/stacks-new-extensible-snapshots

This pull request moves git-dependencies to new file, nbo.yaml (or nbo-front.yaml) that defines a new Stackage snapshot based on lts-9.21 containing our libs.

This way, our repeated builds are a tad bit faster. As seen here,
```
stack build --stack-yaml build-config/backend/stack.yaml --copy-bins --no-run-tests --no-run-benchmarks -j1 --ghc-options=-j1 --ghc-options="+RTS -M3G -RTS" +RTS -N1 -RTS

RSA-2.2.0: unregistering
authenticate-oauth-1.6: unregistering
batch-lib-luna-empire-0.1.0.0: unregistering
batch-lib-request-monitor-0.1.0.0: unregistering
crypto-api-0.13.2: unregistering
entropy-0.3.7: unregistering
lib-ws-connector-0.1.0.0: unregistering
luna-empire-0.1: unregistering
luna-project-0.1: unregistering
luna-stdlib-0.1: unregistering
pureMD5-2.1.3: unregistering
undo-redo-0.1.0.0: unregistering
uuid-1.3.13: unregistering
websockets-0.10.0.0: unregistering
wuss-1.1.3: unregistering
```
for unknown reasons, stack decides to unregister a few of dependencies each time the project is rebuilt, wasting time. I discovered that updating LTS version resolves that.

In addition, passing `--force-dirty` to stack won't trigger a rebuild of our git-dependencies, saving even more time when we decide that rebuilding a project from scratch is necessary.

**Be aware that this is a BREAKING CHANGE in stack.yaml format and requires stack 1.6.1 or newer**

### DIFF
--- a/build-config/backend/nbo.yaml
+++ b/build-config/backend/nbo.yaml
@@ -1,0 +1,77 @@
+resolver: lts-9.21
+name: nbo
+packages:
+- cond-0.4.1.1
+- containers-0.5.8.2
+- error-util-0.0.1.1
+- fingertree-0.1.1.0
+- ghc-mtl-1.2.1.0
+- ghc-session-0.1.2.0
+- path-0.6.1
+- shelly-1.6.8.4
+- serialport-0.4.7
+- git: git@github.com:luna/luna.git
+  commit: 47ca5bca98f45fc96124b789d122fb3553e6056f
+  subdirs:
+  - core
+  - passes
+  - syntax/text/parser
+  - syntax/text/lexer
+  - stdlib
+  - project
+- git: git@github.com:luna/container.git
+  commit: 61ff909efd466735cd4820f9ab2c564e324b5546
+- git: git@github.com:luna/convert.git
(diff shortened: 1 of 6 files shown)


mikusp:
Not much unfortunately. These libs are small, so about 30s on a repeated
build? I have another patch removing building of some abandoned executables
in our repository, but it won't make much of a difference anyway. Still the
biggest time consumers are luna-* packages.

2 lut 2018 16:24 "piotrMocz" <notifications@github.com> napisał(a):

kustosz:
@mikusp looks like everyone accepted the PR, merge?

mikusp:
I guess there's no point in merging it now, considering it will be completely overhauled with new core and its dependencies

PR #795, opened 2018-02-02
Log project settings operations failures.

ljkania:
Let's have some more logging here since it was just needed to track down a problem with old versions of .lunaproject in our packages.

### DIFF
--- a/libs/batch/plugins/luna-empire/src/Empire/Server/Graph.hs
+++ b/libs/batch/plugins/luna-empire/src/Empire/Server/Graph.hs
@@ -173,7 +173,7 @@ getProjectPathAndRelativeModulePath modulePath = do
         absProjectPath <- MaybeT $ findProjectFileForFile absModulePath
         relModulePath  <- MaybeT $ getRelativePathForModule absProjectPath absModulePath
         return (fromAbsFile absProjectPath, fromRelFile relModulePath)
-    when (isNothing mayProjectPathAndRelModulePath) $ Project.logProjectSettingsError "Could not find project path."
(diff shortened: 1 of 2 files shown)


libs/batch/plugins/luna-empire/src/Empire/Server/Graph.hs
```diff
@@ -163,17 +165,19 @@ getDstPortByNodeLoc nl = InPortRef' $ InPortRef nl [Self]
...
-        absModulePath  <- MaybeT . fmap eitherToMaybe . try $ parseAbsFile modulePath
+    let eitherToMaybe :: MonadIO m => Either Path.PathException (Path.Path Path.Abs Path.File) -> m (Maybe (Path.Path Path.Abs Path.File))
+        eitherToMaybe (Left  e) = Project.logProjectSettingsError e >> return def
+        eitherToMaybe (Right a) = return $ Just a
+    mayProjectPathAndRelModulePath <- liftIO . runMaybeT $ do
+        absModulePath  <- MaybeT $ eitherToMaybe =<< try (parseAbsFile modulePath)
         absProjectPath <- MaybeT $ findProjectFileForFile absModulePath
         relModulePath  <- MaybeT $ getRelativePathForModule absProjectPath absModulePath
         return (fromAbsFile absProjectPath, fromRelFile relModulePath)
+    when (isNothing mayProjectPathAndRelModulePath) $ Project.logProjectSettingsError "Could not find project path."
```
mikusp: Probably you intended to use `logProjectPathNotFound` here.

PR #815, opened 2018-02-12
Revise bootstrap luna-develop lib path for zeromq

binary132:
This path is incorrect.

../../libs => ../../lib

### DIFF
--- a/bootstrap.hs
+++ b/bootstrap.hs
@@ -23,7 +23,7 @@ default (T.Text)
 stack = "../../tools/stack/stack"
 
 tools = "../../tools"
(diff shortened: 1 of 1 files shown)


wdanilo:
Actually I'd prefer to change the folder name from `lib` to `libs` - we use `libs` everywhere where additional libraries are unless the fodler contain a single library. @sylwiabr ?

binary132:
This field doesn't do anything on OS X anyway.  Probably the right fix is to use `stack.yaml`: https://github.com/luna/luna-manager/issues/156

Issue #819, opened 2018-02-14
Easier system for custom visualizations


ljkania:
We should provide some system that can discover visualizers defined by user in project.

wdanilo:
What do you mean by it? A "file watcher" ?

ljkania:
In longer perspective, yes. For now we would like to be able at least discover them in folder visualizers as a child of root of project and be able to use them as well.

PR #820, opened 2018-02-14
Use visualizers defined by user in project. #819

ljkania:
First step towards solution. For now user can define their own visualizers in `visualizers` folder which should be a child of project root. For now we "rediscover" them with every reload of graph - in  the (soon) future we would like to have some watcher on this folder instead of current mechanism. Visualizers defined by user appears in visualizers menu with prefix `custom: `


libs/luna-studio-common/src/LunaStudio/Data/NodeValue.hs
```diff
@@ -45,17 +83,18 @@ fromJSVisualizersMap :: Map String (String -> IO String) -> Map VisualizerName V
 fromJSVisualizersMap = Map.fromList . map convertEntry . Map.toList where
     convertEntry (k, v) = (convert k, transformJSVisualizerMatcher v)
 
-applyType :: MonadIO m => TypeRep -> Map VisualizerName VisualizerMatcher -> m (Map VisualizerName VisualizerPath)
+applyType :: MonadIO m => TypeRep -> Map VisualizerId VisualizerMatcher -> m (Map VisualizerId VisualizerPath)
 applyType tpe = fmap (Map.fromList . concat) . liftIO . mapM applyToEntry . Map.toList where
```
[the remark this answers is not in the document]
ljkania: ACK

libs/luna-studio-common/src/LunaStudio/Data/NodeValue.hs
```diff
@@ -32,9 +54,25 @@ instance NFData   VisualizerEntry
...
+
+makeLenses ''Visualizer
+
+instance Binary   Visualizer
+instance NFData   Visualizer
+instance FromJSON Visualizer
+instance ToJSON   Visualizer
+
+type VisualizationData = [Text]
+type VisualizationId   = UUID
```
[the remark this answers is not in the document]
ljkania: Yeah, it seems pretty similar but those are most accurate descriptions... Any ideas @mikusp?

luna-studio/node-editor/src/NodeEditor/State/Global.hs
```diff
@@ -1,4 +1,5 @@
-{-# LANGUAGE DeriveAnyClass #-}
+{-# LANGUAGE DeriveAnyClass    #-}
```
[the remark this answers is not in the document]
wdanilo: DeriveAnyClass is wrong! Never ever use it. Its broken and it does not work. We should write it in our wiki. It makes so many things broken and you will search for errors until end of your life if you ever use it. Youve been warned.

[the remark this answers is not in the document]
mikusp: It has its uses, but without DerivingStrategies it's somewhat unsafe

libs/luna-studio-common/src/LunaStudio/Data/NodeValue.hs
```diff
@@ -45,17 +83,18 @@ fromJSVisualizersMap :: Map String (String -> IO String) -> Map VisualizerName V
...
 
-getMdVis :: MonadIO m => Map VisualizerName VisualizerMatcher -> m (Maybe Visualizer)
-getMdVis visMap = fmap (mdVisName,) . Map.lookup mdVisName <$> applyType (TCons "Text" def) visMap
+getMdVis :: MonadIO m => Map VisualizerId VisualizerMatcher -> m (Maybe Visualizer)
+getMdVis visMap = fmap (Visualizer mdVisId) . Map.lookup mdVisId <$> applyType (TCons "Text" def) visMap
 
-getErrorVis :: MonadIO m => Map VisualizerName VisualizerMatcher -> m (Maybe Visualizer)
-getErrorVis visMap = fmap (errorVisName,) . Map.lookup errorVisName <$> applyType (TCons "Error" def) visMap
+getErrorVis :: MonadIO m => Map VisualizerId VisualizerMatcher -> m (Maybe Visualizer)
+getErrorVis visMap = fmap (Visualizer errorVisId) . Map.lookup errorVisId <$> applyType (TCons "Error" def) visMap
```
kustosz: Won't this break if I define a Luna class named `Error`?

luna-studio/atom/lib/luna-visualizers.coffee
```diff
@@ -15,11 +18,17 @@ normalizeVis = (p, name, visConf) -> (cons) ->
         JSON.stringify(filesToLoad)
     else JSON.stringify(null)
 
-setupConfigMap = (path) ->
-    visualizers = listVisualizers(path)
+setupConfigMap = (projectVisPath) ->
```
kustosz: This is a copy-paste – abstract it into a `getVisualizersForPath` function that returns the resolved visualizers from a given location, then call it twice in this fun.

luna-studio/atom/lib/luna-visualizers.coffee
```diff
@@ -15,11 +18,17 @@ normalizeVis = (p, name, visConf) -> (cons) ->
...
+    visualizers = listVisualizers(visBasePath)
     result = {}
-    result[n] = resolveVis path, n for n in visualizers
-    window.visualizersPath = path
-    window.visualizers     = result
+    result[n] = resolveVis visBasePath, n for n in visualizers
+    window.internalVisualizersPath = visBasePath
```
kustosz: why do you need to store this path in `window`? Is it used afterwards?

luna-studio/node-editor-view/src/NodeEditor/React/View/Sidebar.hs
```diff
@@ -227,8 +228,8 @@ portLabelId = Mount.prefix "focus-portLabel"
 focusPortLabel :: IO ()
 focusPortLabel = UI.focus portLabelId
 
-filterOutSearcherIfNotRelated :: AnyPortRef -> Maybe (Searcher, FilePath) -> Maybe (Searcher, FilePath)
-filterOutSearcherIfNotRelated (OutPortRef' portRef) (Just (s, visLibPath)) = case s ^. Searcher.mode of
-    Searcher.PortName sPortRef _ -> if portRef == sPortRef then Just (s, visLibPath) else Nothing
+filterOutSearcherIfNotRelated :: AnyPortRef -> Maybe (Searcher, VisualizersPaths) -> Maybe (Searcher, VisualizersPaths)
```
kustosz: This is clunky – passing the `VisualizersPaths` around, even though you clearly don't care about it. Are you sure these should be coupled together?

luna-studio/node-editor-view/src/NodeEditor/React/View/Visualization.hs
```diff
@@ -38,12 +42,16 @@ visMenuName     = "visualizers"
 objNameVis      = "node-vis"
 objNameShortVal = "node-short-value"
 
+getVisualizerPath :: VisualizerType -> VisualizersPaths -> Maybe FilePath
+getVisualizerPath (InternalVisualizer) = Just . (^. NE.internalVisualizersPath)
+getVisualizerPath (ProjectVisualizer)  = (^. NE.projectVisualizersPath)
```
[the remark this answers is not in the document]
kustosz: In Luna yes, but this is haskell, operator sections are not that nice :(

luna-studio/node-editor-view/src/NodeEditor/React/View/Visualization.hs
```diff
@@ -80,37 +88,41 @@ docVisualization = React.defineView docViewName $ \(ref, docPresent, visLibPath,
...
-    when (Map.size visualizersMap > 1 || (Map.size visualizersMap == 1 && Map.notMember actVisName visualizersMap)) $ do
-        let menuEntry :: Text -> ReactElementM ViewEventHandler ()
-            menuEntry name = when (name /= actVisName) $
-                li_ [ onClick $ \_ _ -> dispatch ref $ UI.VisualizationEvent $ Visualization.Event visParent $ Visualization.SelectVisualizer visId name ] $ elemString $ convert name
+visualizersMenu :: IsRef r => ReactView (r, VisualizationParent, VisualizationId, VisualizerId, Map VisualizerId VisualizerPath, Bool)
+visualizersMenu = React.defineView visMenuName $ \(ref, visParent, visId, actVisulizerId, visualizersMap, visible) ->
+    when (Map.size visualizersMap > 1 || (Map.size visualizersMap == 1 && Map.notMember actVisulizerId visualizersMap)) $ do
+        let getVisualizerName visualizerId = case visualizerId ^. visualizerType of
+                InternalVisualizer -> visualizerId ^. visualizerName
+                ProjectVisualizer  -> "custom: " <> (visualizerId ^. visualizerName)
```
kustosz: `"project: "` sounds better, everything is `custom` to some extent.

luna-studio/node-editor/src/JS/Visualizers.hs
```diff
@@ -12,17 +12,47 @@ import           GHCJS.Marshal.Pure         (pFromJSVal)
...
 import           LunaStudio.Data.TypeRep    (ConstructorRep, errorTypeRep, toConstructorRep)
 
-foreign import javascript safe "Object.keys(typeof window.visualizers == 'object' ? window.visualizers : {})"
-    getVisualizers' :: IO JSArray
+foreign import javascript safe "window.updateVisualizers($1)"
+    updateVisualizers' :: JSString -> IO ()
 
-getVisualizers :: IO [String]
-getVisualizers = fmap pFromJSVal . toList <$> getVisualizers'
+foreign import javascript safe "window.updateVisualizers()"
```
kustosz: Maybe this function should return the visualizers right away? The dependence on some action writing to `window` and then reading it bothers me. I think we should have a JS API based on functions and the caching behaviors should be it's hidden concern. Even more – the `window.internalVisualizers` could instead be a local variable in the js module, and the only window-exposed stuff would be accessor functions.

luna-studio/node-editor/src/NodeEditor/State/Global.hs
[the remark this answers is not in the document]
wdanilo: @mikusp right, however I dont see any use case for it, which is not covered by safe standalone generic deriving :)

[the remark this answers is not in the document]
wdanilo: btw, the I find the behavior of DerivingStrategies broken by design too :/

luna-studio/atom/lib/luna-visualizers.coffee
wdanilo: There was no place on board for it unfortunatelly @kustosz .

kustosz: WDYM?

luna-studio/node-editor-view/src/NodeEditor/React/View/Visualization.hs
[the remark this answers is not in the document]
kustosz: OH, @wdanilo pointed out that it's about a different set of parens. @mikusp is right.

luna-studio/atom/lib/luna-visualizers.coffee
wdanilo: I mean that if you don't have any free place on a board its much better to write on a window than for example on a wall. It's easier to wash it later etc.

kustosz: Oooooh, I get it now!

luna-studio/node-editor-view/src/NodeEditor/React/View/Visualization.hs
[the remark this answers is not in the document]
mikusp: Operators could be substituted for `view`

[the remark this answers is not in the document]
ljkania: @mikusp, I am not so sure. There is also some function `view` which is used in some of those `.../View/...` modules. It comes from `React.Flux` and I find it less confusing if we stick to `^.` in those files.

libs/luna-studio-common/src/LunaStudio/Data/NodeValue.hs
ljkania: Yup, it will. This design is broken. The correct one is in another pr so once we merged it it won't be a problem anymore.

luna-studio/atom/lib/luna-visualizers.coffee
ljkania: @kustosz - yes. Actually as you can see in this file it is defined there - this is the place where NodeEditor gets this information from. It is needed to correctly target iframe.

luna-studio/node-editor-view/src/NodeEditor/React/View/Sidebar.hs
ljkania: Yup, you're right. I need just first of them. I made a refactor here, and introduced new type for searcher which contains all the information including this path. (FYI: It is not that I do not need them at all. We do not have all state in `View` files so we need to pass them down if they are needed. I do not need `VisualizersPaths` to check if `Searcher` is needed for this port. However at the end of this chain, where we actually use `Searcher`, we need path to md visualizer so we can display documentation correctly).

ljkania:
@pmlodawski, @mikusp, @kustosz: Please review again. There is a lot of nice changes based on CR, but I would feel much more safe if someone go over them.

PR #824, opened 2018-02-15
TC killer

mikusp:
This pull request focuses on providing ability to interrupt running typechecking whenever a new request arrives. Previous approach often resulted in sub-optimal performance and presenting stale results to the user.

On top of that, general robustness of typechecker is improved and there should be fewer situations when observed behaviour is not what was expected.


libs/batch/plugins/luna-empire/src/Empire/Server.hs
```diff
@@ -115,26 +125,44 @@ prepareStdlib = do
     (cleanup, std) <- Typecheck.createStdlib $ lunaroot <> "/Std/"
     return (std, cleanup)
 
-startTCWorker :: Empire.CommunicationEnv -> Bus ()
-startTCWorker env = liftIO $ do
+killPreviousTC :: Empire.CommunicationEnv -> Maybe (Async Empire.InterpreterEnv) -> IO ()
+killPreviousTC env prevAsync = case prevAsync of
+    Just a -> do
+        res <- Async.poll a
+        case res of
```
ljkania: `Just a -> Async.poll a >>= \case` could save you extra indent and temp variable.

libs/batch/plugins/luna-empire/src/Empire/Server/Graph.hs
```diff
@@ -126,7 +126,7 @@ import qualified ZMQ.Bus.Data.Message                    as Message
 import qualified ZMQ.Bus.EndPoint                        as EP
 import           ZMQ.Bus.Trans                           (BusT (..))
 import qualified ZMQ.Bus.Trans                           as BusT
-
+import qualified System.IO as IO
```
ljkania: Debug leftover?

libs/luna-empire/src/Empire/Commands/Typecheck.hs
```diff
@@ -44,13 +56,17 @@ import qualified Luna.Project                     as Project
...
+import           Luna.Syntax.Text.Parser.Errors   (Invalids)
+import           OCI.IR.Layout.Typed              (type (>>))
 import           OCI.IR.Name.Qualified            (QualName)
-
-import           System.Directory                     (canonicalizePath)
-import           System.Environment                   (getEnv)
-import qualified Path
+import qualified OCI.Pass.Class                   as Pass (SubPass, eval')
+import qualified OCI.Pass.Manager                 as Pass (PassManager, setAttr, State)
+import           System.Log                                   (DropLogger(..), dropLogs)
```
ljkania: DITTO

libs/luna-empire/src/Empire/Commands/Typecheck.hs
```diff
@@ -1,33 +1,43 @@
...
-import           Control.Monad.State              (execStateT)
+import           Control.Monad.State              (get, execStateT, runStateT)
 import qualified Data.Map                         as Map
 import           Data.Maybe                       (maybeToList)
+import qualified Data.Set                         as Set
 import           Empire.Prelude                   hiding (toList)
 import           Prologue                         (catMaybes, mapping)
 import           System.Directory                 (withCurrentDirectory)
+import           System.Directory                     (canonicalizePath)
+import           System.Environment                   (getEnv)
```
ljkania: Could you fix the indentation?

libs/luna-empire/src/Empire/Commands/Typecheck.hs
```diff
@@ -68,35 +84,37 @@ runInterpreter path imports = runASTOp $ do
...
             res     <- Interpreter.interpret' imports . IR.unsafeGeneralize $ bodyRef
-            result  <- liftIO $ withCurrentDirectory (maybe (takeDirectory path) Path.toFilePath rootPath) $ runIO $ runError $ execStateT res def
-            case result of
-                Left e  -> return Nothing
-                Right r -> return $ Just r
+            mask_ $ do
+                result  <- liftIO $ withCurrentDirectory (maybe (takeDirectory path) Path.toFilePath rootPath) $ runIO $ runError $ execStateT res def
+                case result of
+                    Left e  -> return Nothing
+                    Right r -> return $ Just r
```
ljkania: `return $ either Nothing (Just r) result` ?

libs/luna-empire/src/Empire/Commands/Typecheck.hs
```diff
@@ -168,26 +187,31 @@ getCurrentScope imports file = do
...
-            scope      <- (if recompute then recomputeCurrentScope else getCurrentScope) imports file
-            let imps = unionImports (flattenScope $ Scope std) scope
-            zoom graph $ flip (zoomBreadcrumb' br) (return ()) $ do
-                runTC imps
-                updateNodes  loc
-                {-updateMonads loc-}
-                when interpret $ do
-                    scope <- runInterpreter file imps
-                    traverse_ (updateValues loc) scope
+            void $ mask_ $ recomputeCurrentScope imports file
```
ljkania: It does not need `do`, does it?

libs/luna-empire/src/Empire/Commands/Typecheck.hs
mikusp: @kustosz I see we're ignoring the exception string here, was it done on purpose?

libs/luna-empire/src/Empire/Commands/Typecheck.hs
mikusp: While technically it doesn't, I think it looks ok here to have consistency with next clause

libs/batch/plugins/luna-empire/src/Empire/Server.hs
mikusp: done

libs/batch/plugins/luna-empire/src/Empire/Server/Graph.hs
mikusp: yup

libs/batch/plugins/luna-empire/src/Empire/Server/Graph.hs
```diff
@@ -487,9 +486,15 @@ handleSubstitute = modifyGraph defInverse action replyResult where
     action req@(Substitute.Request location diffs) = do
         let file = location ^. GraphLocation.filePath
         prevImports <- Graph.getAvailableImports location
-        res         <- withDefaultResultTC location $ Graph.substituteCodeFromPoints file diffs
+        res         <- withDefaultResult location $ Graph.substituteCodeFromPoints file diffs
         newImports  <- Graph.getAvailableImports location
-        return $ Substitute.Result res $ if Set.fromList prevImports == Set.fromList newImports then def else return newImports
+        let importChange = if Set.fromList prevImports == Set.fromList newImports then Nothing else Just newImports
+        if isJust importChange then do
+            Graph.typecheckWithRecompute (GraphLocation file def)
```
kustosz: Do we really need both? The whole point of `typecheckWithRecompute` was to run exactly the operation of running toplevel TC, so why is this needed?

libs/luna-empire/src/Empire/ASTOp.hs
```diff
@@ -245,36 +252,51 @@ runTypecheck imports = do
...
+
+tcInit = do
+    Pass.setAttr (getTypeDesc @WorldExpr)                 $ error "Data not provided: WorldExpr"
+    Pass.setAttr (getTypeDesc @UnitLoader.UnitsToLoad)    $ error "Data not provided: UnitsToLoad"
+    Pass.setAttr (getTypeDesc @UnitLoader.SourcesManager) $ error "Data not provided: SourcesManager"
+    Pass.setAttr (getTypeDesc @UnitSet)                   $ error "Data not provided: UnitSet"
+    Pass.setAttr (getTypeDesc @Invalids)                  $ (mempty :: Invalids)
+    initNameGen
+
+extractImportedModules unit = do
```
kustosz: Type it, no cheating!

libs/luna-empire/src/Empire/ASTOp.hs
```diff
@@ -245,36 +252,51 @@ runTypecheck imports = do
...
     put $ newG & Graph.ast .~ AST st passSt
 
+evalTC g ir pmState = flip runStateT g
+                    . withVis
+                    . dropLogs
+                    . DepState.evalDefStateT @Cache
+                    . flip evalIRBuilder ir
+                    . flip evalPassManager pmState
+
+tcInit = do
```
kustosz: Type it, no cheating!

libs/luna-empire/src/Empire/Commands/Graph.hs
```diff
@@ -205,6 +206,8 @@ addImports loc@(GraphLocation file _) modulesToImport = do
         let newImports = map (\i -> Text.concat ["import ", i, "\n"]) neededImports
         return $ Text.concat $ newImports ++ [code]
     reloadCode loc newCode
+    typecheckWithRecompute (GraphLocation file def)
```
kustosz: again, why recompute after recompute?

libs/luna-empire/src/Empire/Commands/Typecheck.hs
kustosz: It was done because I had no better idea what to do with this error. This says "the whole computation ended with an exception", without even a hint as to what node caused it. At some point it will be probably displayed somewhere, but is ignored for a long time now.

libs/batch/plugins/luna-empire/src/Empire/Server.hs
```diff
@@ -115,26 +126,42 @@ prepareStdlib = do
     (cleanup, std) <- Typecheck.createStdlib $ lunaroot <> "/Std/"
     return (std, cleanup)
 
-startTCWorker :: Empire.CommunicationEnv -> Bus ()
-startTCWorker env = liftIO $ do
+killPreviousTC :: Empire.CommunicationEnv -> Maybe (Async Empire.InterpreterEnv) -> IO ()
+killPreviousTC env prevAsync = case prevAsync of
+    Just a -> Async.poll a >>= \case
```
piotrMocz: Actually, you could simplify it even further by writing:
```
killPreviousTC env = \case
    Nothing -> return ()
    Just prevAsync -> Async.poll prevAsync >>= \case [...]
```
Oh and I like to put the significantly-shorter branch first, so that you don't get a lone `_ -> return ()` at the end of the function. Just a matter of preference, though.

libs/luna-empire/src/Empire/ASTOp.hs
```diff
@@ -245,36 +253,58 @@ runTypecheck imports = do
         return (st, passSt)
     put $ newG & Graph.ast .~ AST st passSt
 
+evalTC :: ClsGraph
+       -> IR
+       -> State (Pass.PassManager (IRBuilder (DepState.StateT Cache (Logger DropLogger (Vis.VisStateT (StateT ClsGraph IO))))))
+       -> Pass.PassManager (IRBuilder (DepState.StateT Cache (Logger DropLogger (Vis.VisStateT (StateT ClsGraph IO))))) a
```
piotrMocz: o.0

PR #850, opened 2018-03-02
Show compile error backtraces as visualizations #849

mikusp:
Compile errors should look like this

zrzut ekranu z 2018-03-02 12-58-31

### DIFF
--- a/libs/luna-empire/src/Empire/Commands/Graph.hs
+++ b/libs/luna-empire/src/Empire/Commands/Graph.hs
@@ -213,7 +213,7 @@ addImports loc@(GraphLocation file _) modulesToImport = do
         modulesMVar <- view modules
         importPaths <- liftIO $ getImportPaths loc
         Lifted.modifyMVar modulesMVar $ \cmpModules -> do
-            res     <- runModuleTypecheck qualName importPaths cmpModules
+            res <- runModuleTypecheck qualName importPaths cmpModules
             case res of
                 Left err                          -> liftIO (print err) >> return (cmpModules, ())
                 Right (newImports, newCmpModules) -> return (newCmpModules, ())
--- a/libs/luna-empire/src/Empire/Commands/Graph.hs
+++ b/libs/luna-empire/src/Empire/Commands/Graph.hs
@@ -136,6 +136,7 @@ import qualified Empire.Commands.GraphBuilder     as GraphBuilder
 import qualified Empire.Commands.GraphUtils       as GraphUtils
 import qualified Empire.Commands.Library          as Library
 import qualified Empire.Commands.Publisher        as Publisher
+import qualified Empire.Commands.Typecheck        as Typecheck
 import           Empire.Data.AST                  (InvalidConnectionException (..), EdgeRef, NodeRef, NotInputEdgeException (..),
                                                    NotUnifyException, PortDoesNotExistException(..), ConnectionException(..),
                                                    SomeASTException, astExceptionFromException, astExceptionToException)
@@ -207,11 +208,12 @@ addImports loc@(GraphLocation file _) modulesToImport = do
         return $ Text.concat $ newImports ++ [code]
     reloadCode loc newCode
     typecheckWithRecompute loc
+    qualName <- Typecheck.filePathToQualName file
     withUnit (GraphLocation file def) $ do
         modulesMVar <- view modules
         importPaths <- liftIO $ getImportPaths loc
         Lifted.modifyMVar modulesMVar $ \cmpModules -> do
-            res     <- runModuleTypecheck importPaths cmpModules
+            res     <- runModuleTypecheck qualName importPaths cmpModules
             case res of
                 Left err                          -> liftIO (print err) >> return (cmpModules, ())
                 Right (newImports, newCmpModules) -> return (newCmpModules, ())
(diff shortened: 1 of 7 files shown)


mikusp:
@ljkania any hints how to achieve that?

libs/luna-empire/src/Empire/Commands/Graph.hs
```diff
@@ -207,11 +208,12 @@ addImports loc@(GraphLocation file _) modulesToImport = do
         return $ Text.concat $ newImports ++ [code]
     reloadCode loc newCode
     typecheckWithRecompute loc
+    qualName <- Typecheck.filePathToQualName file
     withUnit (GraphLocation file def) $ do
         modulesMVar <- view modules
         importPaths <- liftIO $ getImportPaths loc
         Lifted.modifyMVar modulesMVar $ \cmpModules -> do
-            res     <- runModuleTypecheck importPaths cmpModules
+            res     <- runModuleTypecheck qualName importPaths cmpModules
```
ljkania: Does this arrow correspond anyhow with the previous ones? If not - why space?

libs/luna-empire/src/Empire/Commands/Typecheck.hs
```diff
@@ -90,7 +90,11 @@ updateNodes loc@(GraphLocation _ br) = do
             errs <- IR.getLayer @IR.Errors =<< ASTRead.getASTRef nid
             case errs of
                 []     -> return Nothing
-                e : es -> return $ Just $ (nid, NodeError $ APIError.Error APIError.CompileError $ e ^. Errors.description)
+                e : es -> do
+                    let toSrcLoc (Errors.ModuleTagged mod (Errors.FromMethod klass method)) = APIError.SourceLocation (convert mod) (Just (convert klass)) (convert method)
+                        toSrcLoc (Errors.ModuleTagged mod (Errors.FromFunction function))   = APIError.SourceLocation (convert mod) Nothing (convert function)
+                    let errorDetails = APIError.CompileErrorDetails (map toSrcLoc (e ^. Errors.arisingFrom)) (map toSrcLoc (e ^. Errors.requiredBy))
```
ljkania: No need for another `let`

libs/luna-studio-common/src/LunaStudio/Data/Error.hs
```diff
@@ -5,22 +5,44 @@ import           Data.Aeson.Types (FromJSON, ToJSON)
...
 import           Prologue
 
-data NodeError  = CompileError | RuntimeError deriving (Eq, Generic, Show)
-data GraphError = BreadcrumbDoesNotExist | ParseError | OtherGraphError deriving (Eq, Generic, Show)
-data LunaError  = Graph GraphError | OtherLunaError deriving (Eq, Generic, Show)
+data NodeError  = CompileError CompileErrorDetails | RuntimeError deriving (Generic, Show)
+data GraphError = BreadcrumbDoesNotExist | ParseError | OtherGraphError deriving (Generic, Show)
+data LunaError  = Graph GraphError | OtherLunaError deriving (Generic, Show)
+
+data CompileErrorDetails = CompileErrorDetails { _arisingFrom :: [SourceLocation]
```
ljkania: I know that our code does not follow what @wdanilo wrote down, but we should start it right now. So brackets should be in the next line.

libs/luna-studio-common/src/LunaStudio/Data/Error.hs
```diff
@@ -5,22 +5,44 @@ import           Data.Aeson.Types (FromJSON, ToJSON)
...
+                                               , _requiredBy  :: [SourceLocation]
+                                               } deriving (Generic, Show)
+
+data SourceLocation = SourceLocation { _mod   :: Text
```
ljkania: ditto

mikusp:
Please recheck, I submitted the requested fixes

kustosz:
There are some issues with the JS:
1. Inconsistent code formatting – some blocks start on the same line, some opening braces get their own line. The whole file should follow:
```
if (...) {
} else if (...) {
} else {
}
```
2. The use of `<br />` is wrong – the lines you separate with it have different semantics and that should be mirrored in the document structure. Instead of things like `foo <br /> bar` use `<p>foo</p><p>bar</p>` or `<h3>foo</h3><p>bar</p>` or any other combination of headers, parahraphs, lists or whatever makes sense from the semantic perspective.
3. The all-encompassing `<pre>` tag is probably not needed – we are not depending on whitespace-preformatted text anymore. If it used just for the default monospace styling, remove it and use CSS.
4. There's no need for mutable vars like `requiredBy` – JS arrays have a `reduce` method, which is basically `fold`. And the fact you had to use single-field object just to get mutation to work suggests the flow is too complicated.
5. It is better to construct the whole document contents and then assign than construct it gradually in-place – each assignment to `document.body.innerHTML` results in a redraw, and most of the time it's not even valid HTML.

ljkania:
screen shot 2018-03-06 at 13 25 19
Is this correct?

kustosz:
@mikusp regarding issue from @ljkania – I think the current function is being compiled as `TgtNone`. If you set it to `TgtDef` with proper name, it should work. https://github.com/luna/luna-studio/blob/6b1754e5e57a965ccae7596ff9d4038bd4d7638b/libs/luna-empire/src/Empire/ASTOp.hs#L250

PR #855, opened 2018-03-08
Add searcher to basegl gui

pmlodawski:
It is still missing some features:
- mock node is not displayed
- input is not displaying correctly
- documentation is not displaying

Added styles are copied from atom version and should be reviewed.


luna-studio/node-editor-view/src/NodeEditor/View/Searcher.hs
```diff
@@ -0,0 +1,80 @@
...
+    , selected :: Int
+    , entries  :: [EntryView]
+    , input    :: String
+    } deriving (Generic, Show)
+
+instance ToJSON HighlightView
+instance ToJSON EntryView
+instance ToJSON SearcherView
+
+instance Convertible (Int, Int) HighlightView where
```
ljkania: Really need this convert? It is simply `uncurry HighlightView`

luna-studio/node-editor-basegl/src/view/Searcher.coffee
```diff
@@ -1,17 +1,85 @@
...
+                prefixSpan.className = style.luna ['searcher__pre']
+                prefixSpan.innerHTML = entry.className or ''
+                resultName.appendChild prefixSpan
+
+                nameSpan = document.createElement 'span'
+                resultName.appendChild nameSpan
+
+                name = entry.name
+                last = 0
+                if entry.highlights instanceof Array
```
kustosz: This may indicate something wrong. Checking a type this deep into the logic suggests that there is something wrong with the data model – it should be well-formed at the system boundary. E.g. can we make it so that `entry.highlights` is always an array, just empty sometimes?
This deep into the logic, you should be sure that all data conforms to specified contracts and should not explicitly check for types.

luna-studio/node-editor-view/src/NodeEditor/View/SidebarNode.hs
```diff
@@ -21,19 +21,23 @@ outputNodeView new old =
...
+        , inPorts  :: [PortView]
+        , outPorts :: [PortView]
         } deriving (Generic, Show)
 
 instance ToJSON SidebarNodeView
 instance Convertible InputNode SidebarNodeView where
     convert n = SidebarNodeView
-        {- ports -} def
-        {- ports -} (n ^. SidebarNode.inputMode . to show)
+        {- key      -} (n ^. SidebarNode.nodeLoc . to show)
```
kustosz: Haskell has syntax for this:
```
SidebarNodeView { key = (...), inPorts = def, outPorts = (...) }
```
If you feel like describing it in a comment, might as well describe it in code.

luna-studio/node-editor/src/NodeEditor/Event/Processor.hs
```diff
@@ -94,8 +94,9 @@ handleAnyException :: SomeException -> IO ()
 handleAnyException = error . show
 
 handleExcept :: State -> Event -> JSException -> IO State
-handleExcept oldState event except = do
-    error $ "JavaScriptException: " <> show except <> "\n\nwhile processing: " <> show event
+handleExcept oldState event except@(JSException jsval _) = do
+    consoleLog' jsval
+    error $ show except <> "\n\nwhile processing: " <> show event
```
kustosz: Is this a temporary measure for debugging?

luna-studio/text-editor/src/TextEditor/Event/Processor.hs
```diff
@@ -54,8 +54,9 @@ connectEventSources conn loop = do
     sequence_ $ mkSource <$> handlers
 
 handleExcept :: State -> Event -> JSException -> IO State
-handleExcept oldState event except = do
-    putStrLn $ "JavaScriptException: " <> show except <> "\n\nwhile processing: " <> show event
+handleExcept oldState event except@(JSException jsval _) = do
```
ljkania: This function is exact duplicate.

luna-studio/node-editor-view/src/NodeEditor/View/Searcher.hs
```diff
@@ -0,0 +1,80 @@
...
+    convert (start, end) = HighlightView
```
kustosz: same here, use record syntax instead of comments

luna-studio/atom/lib/luna-node-editor-tab.coffee
```diff
@@ -19,8 +19,9 @@ class LunaNodeEditorTab extends View
         pushShortcutEvent = (name, arg = null) => @nodeEditor.pushEvent({_shortcut: name, _arg : arg})
         pushSearcherEvent = (name, arg = null) => @nodeEditor.pushEvent(if arg == null then {tag: name} else {tag: name, contents : arg})
         @nodeEditor.start @uri, mountPoint
-        window.install = => nodeEditorBaseGL.install mountPoint2, @nodeEditor.setView
-        window.nodeEditor = nodeEditorBaseGL
+        window.install = => nodeEditorBaseGL.install mountPoint2, 'atom://luna-studio/rsc/',  (ne) =>
+            window.nodeEditor = ne
```
wdanilo: If this is not only for debug purposes and will not be deleted soon, we should definitely keep such things under common name, like `window.luna.nodeEditor`

luna-studio/node-editor-basegl/dist/index.html
```diff
@@ -9,6 +9,20 @@
...
+		<link rel="stylesheet/less" type="text/css" href="style/_connections.less" />
+		<!-- <link rel="stylesheet/less" type="text/css" href="style/_general.less" /> -->
+		<link rel="stylesheet/less" type="text/css" href="style/_mixins.less" />
+		<!-- <link rel="stylesheet/less" type="text/css" href="style/_nodes.less" /> -->
+		<link rel="stylesheet/less" type="text/css" href="style/_searcher.less" />
+		<link rel="stylesheet/less" type="text/css" href="style/_selection.less" />
+		<!-- <link rel="stylesheet/less" type="text/css" href="style/_sideports.less" /> -->
+		<link rel="stylesheet/less" type="text/css" href="style/_temp.less" />
+		<link rel="stylesheet/less" type="text/css" href="style/_variables-global.less" />
+		<link rel="stylesheet/less" type="text/css" href="style/_variables.less" />
```
wdanilo: Why don't you create a single less file which imports other less files? It's both prettier and faster to compute (less caches results between compilation of these files)

wdanilo: one more thing - why our styles start with underscore?

luna-studio/node-editor-view/src/NodeEditor/View/Searcher.hs
wdanilo: :+1:

luna-studio/node-editor-view/src/NodeEditor/View/Searcher.hs
```diff
@@ -0,0 +1,80 @@
...
+
+entries' :: Searcher.Mode -> [EntryView]
+entries' = \case
+    Searcher.Command    m -> convert m
+    Searcher.Node   _ _ m -> convert m
+    Searcher.NodeName _ m -> convert m
+    Searcher.PortName _ m -> convert m
+
+foreign import javascript safe "atomCallback.getNodeEditorView().setSearcher($1)"
+    setSearcher' :: JSVal -> IO ()
```
wdanilo: `setSearcher'` is an internal function I suppose. If so, I'd strongly prefer to name it `setSearcher__`. The apostrophe on the end doesnt suggest its internal function and basically is allowed for normal function names.

luna-studio/node-editor-view/src/NodeEditor/View/NodeEditor.hs
```diff
@@ -1,15 +1,17 @@
...
+import           NodeEditor.View.Searcher          (searcherView)
+import           NodeEditor.View.SidebarNode       (inputNodeView, outputNodeView)
 
 
 nodeEditorView :: MonadIO m => NodeEditor -> NodeEditor -> m ()
 nodeEditorView new old = do
     expressionNodesView (new ^. expressionNodes) (old ^. expressionNodes)
+    inputNodeView   (new ^. inputNode)   (old ^. inputNode)
+    outputNodeView  (new ^. outputNode)  (old ^. outputNode)
+    searcherView    (new ^. searcher)    (old ^. searcher)
```
wdanilo: Why not 
```
let onBoth = on (,) 
inputNodeView $ onBoth (view inputNode) new old
```
or any further refactoring? This way you dont have to type the field name twice!

luna-studio/node-editor-view/src/NodeEditor/View/Searcher.hs
```diff
@@ -0,0 +1,80 @@
...
+import           NodeEditor.React.Model.Searcher (Searcher)
+import qualified NodeEditor.React.Model.Searcher as Searcher
+import           LunaStudio.Data.NodeSearcher    (Match)
+import qualified LunaStudio.Data.NodeSearcher    as Match
+
+
+
+searcherView :: MonadIO m => Maybe Searcher -> Maybe Searcher -> m ()
+searcherView new old =
+    when (new /= old) $ setSearcher $ convert new
```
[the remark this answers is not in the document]
wdanilo: According to linter, it should be written as `when (new /= old) . setSearcher $ convert new`

luna-studio/node-editor-view/src/NodeEditor/View/Searcher.hs
```diff
@@ -0,0 +1,80 @@
...
+    Searcher.Node nl _ _   -> Just $ show nl
+    Searcher.NodeName nl _ -> Just $ show nl
+    _                      -> Nothing
+
+entries' :: Searcher.Mode -> [EntryView]
+entries' = \case
+    Searcher.Command    m -> convert m
+    Searcher.Node   _ _ m -> convert m
+    Searcher.NodeName _ m -> convert m
+    Searcher.PortName _ m -> convert m
```
wdanilo: First of all, the `entries` function is an field accessor of `Searcher.Mode` and should be placed in searcher module. Second, every `Searcher.Mode` constructor has this field, it should be refactored to be nested datatype then!

luna-studio/node-editor-view/src/NodeEditor/View/Searcher.hs
wdanilo: The same applies to other files, like here: https://github.com/luna/luna-studio/blob/32bcc193bb6762cf8bf5bdfdbb3af3a052cb4e28/luna-studio/node-editor-view/src/NodeEditor/React/Model/NodeEditor.hs

```
instance Default NodeEditor where
    def = NodeEditor
        {- expressionNodes          -} def
        {- inputNode                -} def
        {- outputNode               -} def
        {- monads                   -} def
        {- connections              -} def
        {- visualizersLibPath       -} def
        {- visualizations           -} def
        {- visualizationsBackup     -} def
        {- halfConnections          -} def
        {- connectionPen            -} def
        {- selectionBox             -} def
        {- searcher                 -} def
        {- textControlEditedPortRef -} def
        {- graphStatus              -} NoGraph
        {- layout                   -} def
        {- topZIndex                -} def
```

luna-studio/node-editor-view/src/NodeEditor/View/Searcher.hs
```diff
@@ -0,0 +1,80 @@
...
+    , className  :: String
+    , highlights :: [HighlightView]
+    } deriving (Generic, Show)
+
+data SearcherView = SearcherView
+    { key      :: Maybe String
+    , selected :: Int
+    , entries  :: [EntryView]
+    , input    :: String
+    } deriving (Generic, Show)
```
wdanilo: Why arent you using Lenses here? We agreed to use Lenses on all single constructor datatypes. We have even prepared special ToJSON / FromJSON wrappers to handle underscore names correctly.

luna-studio/node-editor-basegl/src/view/Searcher.coffee
```diff
@@ -1,17 +1,85 @@
...
+            root.id = searcherRoot
+            root.style.width = 100 + 'px'
+            root.style.height = 200 + 'px'
+            @def = basegl.symbol root
 
     updateView: =>
-        @view.position.xy = @parent.node(@key).position
+        node = @parent.node(@key)
+        if node?
+            @view.position.xy = node.position
```
wdanilo: thy not single line here? 
`if node? then @view.position.xy = node.position`

luna-studio/atom/lib/luna-node-editor-tab.coffee
pmlodawski: It is debug only, I'm going to remove all window bound variables before merging to master

luna-studio/node-editor-basegl/dist/index.html
pmlodawski: Single file is a good idea, I will change it. 

Why it starts with underscores? I don't know, I simply copied it from atom, we should ask @janklosowski, why he decide to start them with underscore

pmlodawski: I guess it is somehow connected with the fact that atom load files alphabetically. For now, I wont change it to keep track which files was copied from atom and which are new.

luna-studio/node-editor/src/NodeEditor/Event/Processor.hs
mikusp: Do not use name `error` if you don't mean `Prelude.error`, it's confusing

luna-studio/node-editor-view/src/NodeEditor/View/NodeEditor.hs
pmlodawski: good point, thanks! I refactored it to 
```
nodeEditorView :: MonadIO m => DiffT NodeEditor m ()
nodeEditorView = do
    diff expressionNodesView expressionNodes
    diff inputNodeView       inputNode
    diff outputNodeView      outputNode
    diff searcherView        searcher
    diff connectionsView     connections
```

luna-studio/node-editor-view/src/NodeEditor/View/Searcher.hs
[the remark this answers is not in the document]
pmlodawski: Refactored it to 
```
searcherView :: MonadIO m => DiffT (Maybe Searcher) m ()
searcherView = diffApply $ setSearcher . convert
```

luna-studio/node-editor-view/src/NodeEditor/View/Searcher.hs
pmlodawski: I think maybe I should create own Convert with different semantics: extracting required visual elements for gui model

luna-studio/node-editor-view/src/NodeEditor/View/Searcher.hs
ljkania: It may actually be leftover of my work - I was no aware of those special wrappers. How do you use them?

luna-studio/node-editor-view/src/NodeEditor/View/Searcher.hs
ljkania: @wdanilo - If you need to - please make an issue, but there is no good in pointing files unchanged by pr.

luna-studio/node-editor-view/src/NodeEditor/View/Searcher.hs
ljkania: Also my fault - however I think that we should keep refactor of this as a separate pr. Should we make an issue for this refactor?

luna-studio/node-editor-view/src/NodeEditor/View/Searcher.hs
ljkania: @wdanilo: This is format is used across luna-studio for foreign imports functions - maybe we need another refactor here?

luna-studio/node-editor-basegl/src/view/Searcher.coffee
ljkania: Too haskelly for me :) I prefer format @pmlodawski used here. Assignment is very specific for me.
But why not: `@view.position.xy = node?.position` - this should work fine, right @kustosz?

luna-studio/text-editor/src/TextEditor/Event/Processor.hs
pmlodawski: Yes, it is. The whole file is duplicate, since text-editor was copied from node-editor.

luna-studio/node-editor-view/src/NodeEditor/View/Searcher.hs
wdanilo: @pmlodawski What such convert would do? Basically I tend to reuse this typeclass whenever possible, for example:
```
type FromBool a = Convertible Bool a
fromBool :: FromBool a => Bool -> a
fromBool = convert
```

It just gives you much more powerful base, but allows still for every operation you want. Remember that convert handles a lot of important cases. You can for example use `convert'` to allow any conversion or `convert` to disallow `id` convertion. This is very helpful for discovering unused converts - if you change api and try to `convert` from `String` to `String` you get error.

luna-studio/atom/lib/luna-node-editor-tab.coffee
wdanilo: Im not convinced if you shoudl remove them. Id rather create a global flag, which when set, will not remove them,. This way we could set a global DEBUG flag and booom have access to all things from browser.

luna-studio/node-editor-view/src/NodeEditor/View/Searcher.hs
wdanilo: If it will not be merged here, then yes, we need to document it somewhere.

luna-studio/node-editor-view/src/NodeEditor/View/Searcher.hs
wdanilo: they are in `lens-utils/Control.Lens.Aeson` and you use them exaclty like `Aeson docs tell you to use them:

```
import qualified Control.Lens.Aeson as Lens
instance ToJSON Person where
    toEncoding = Lens.toEncoding```

File: luna-studio/node-editor-view/src/NodeEditor/View/Searcher.hs (hunk 12 above)
Comment (ljkania) [reply luna-studio/node-editor-view/src/NodeEditor/View/Searcher.hs]: @pmlodawski: I'll let you decide on this one :) If you decide to leave it for another time - please make an issue and assign me

File: luna-studio/node-editor-view/src/NodeEditor/View/Searcher.hs (hunk 1 above)
Comment (pmlodawski) [reply luna-studio/node-editor-view/src/NodeEditor/View/Searcher.hs]: @wdanilo, you are right, there is no point to do this.

File: luna-studio/node-editor-basegl/src/view/Searcher.coffee [hunk 15]
```diff
@@ -1,17 +1,85 @@
...
+                prefixSpan = document.createElement 'span'
+                prefixSpan.className = style.luna ['searcher__pre']
+                prefixSpan.innerHTML = entry.className or ''
+                resultName.appendChild prefixSpan
+
+                nameSpan = document.createElement 'span'
+                resultName.appendChild nameSpan
+
+                name = entry.name
+                last = 0
```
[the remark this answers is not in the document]
pmlodawski: You are right, this part was copied from recursive haskell code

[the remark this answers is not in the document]
pmlodawski: This is not only about getting last element of array, but it stores information about where rendering end in previous step

luna-studio/node-editor-view/src/NodeEditor/View/SidebarNode.hs
pmlodawski: Using this syntax will not raise compile error if you add new field to `SidebarNodeView` but will leave this field undefined. This is because I'm not using it.

luna-studio/node-editor/src/NodeEditor/Event/Processor.hs
pmlodawski: @kustosz yes and no. `Show JSException` shows only error, but no stack trace and because of that its very hard to find errors in js code. 
This `consoleLog'` function which I added, calls `console.log` on `JSVal` which prints stack trace, but I have no idea, how can I get this stack trace from `JSException` as `String` so I could display it in message box and remove this temporary solution

@mikusp we should change it on `master` first, not on this branch

luna-studio/node-editor-view/src/NodeEditor/View/Searcher.hs
pmlodawski: @kustosz @wdanilo record syntax for initialization is dangerous, we should never use it. It may leave fields undefined without raising compile error.

luna-studio/node-editor/src/NodeEditor/Event/Processor.hs
kustosz: @pmlodawski yeah, I was convinced this is `Prelude.error`, hence my question. If that's something else then it needs changing rather soon (not saying now).

luna-studio/node-editor-view/src/NodeEditor/View/Searcher.hs
pmlodawski: @wdanilo @ljkania I agree with both of you. This should definitely be refactored in `Searcher.Mode` but not on this branch. I'm trying not to touch any luna-studio files if it is not necessary. I'll make a separate pull request for this and then change it here.

luna-studio/node-editor/src/NodeEditor/Event/Processor.hs
pmlodawski: I'll make a PR for that, since I rememeber that this confusing `error` was mentioned some time before but nothing was done about it

luna-studio/node-editor-basegl/src/view/Searcher.coffee
pmlodawski: Good point, thanks!

luna-studio/node-editor-view/src/NodeEditor/View/Searcher.hs
pmlodawski: I just found that there exists `hints` lens, which is exactly what I need here.

luna-studio/node-editor-view/src/NodeEditor/View/SidebarNode.hs
kustosz: Didn't know that, that is disgusting, shame on you, Haskell :/

luna-studio/node-editor-view/src/NodeEditor/View/Searcher.hs
pmlodawski: @wdanilo, you meant
```haskell
import qualified Control.Lens.Aeson as Lens
instance ToJSON Person where
    toEncoding = Lens.toEncoding
    toJSON     = Lens.toJSON
```

luna-studio/node-editor-basegl/src/view/Searcher.coffee
pmlodawski: @ljkania, just noticed that it translates to 
```js
this.view.position.xy = typeof node !== "undefined" && node !== null ? node.position : void 0;
```

luna-studio/node-editor-basegl/dist/index.html
Spaceghost: The underscores, if I recall, denote being mix-ins or partials, not complete stylesheets on their own.

pmlodawski: That make sense

PR #856, opened 2018-03-09
add random part to tmp folder

sylwiabr:


runner/src/StudioRunner.hs
```diff
@@ -66,7 +68,7 @@ data RunnerConfig = RunnerConfig { _versionFile            :: FilePath
                                  , _supervisorKillFolder   :: FilePath
                                  , _supervisorKillBin      :: FilePath
                                  , _atomBinPath            :: FilePath
-                                 , _mainTmpDirectory       :: FilePath
+                                 , _mainTmpDirectory       :: T.Text
```
piotrMocz: Why are you changing that to `Text`? You're running `fromText` on that later (and never actually using it as `Text`). FWIW, I would stick with `FilePath`.

PR #915, opened 2018-06-03
WS rework

kustosz:
Okay, soo... It turns out our websockets server sucked big time. At this point I'm not sure if that's only on macs or were linuxes affected too. For macs, any visualization result of ~1MB had a 50% chance of never making it to GUI and anything larger just never made it. Anyway, it now works better on both platforms (i.e. it actually started working on mac and is faster on linux). What happened in this PR:

1. I have reworked the sockets to use binary data instead of base64-encoding everything. Reduces encoding overhead and reduces package sizes by a factor of 2.
2. Swapped STM's TChans for Unagi chans. This fixed some memory leaks and, in particular a lot of busy spin-locking. Slap me if I ever want to use STM again.
3. This is what makes macOS distribution work. We are changing the implementation from a single, duplex websocket to two separate connections, one for each direction of GUI <-> backend communications. "Why", you may ask, "aren't websockets duplexed by default and should work just fine in this mode?". They should. But the HS `websockets` library has a rather serious bug that breaks duplex. We may switch back to a single socket once this issue is resolved: https://github.com/jaspervdj/websockets/issues/178.

### DIFF
--- a/libs/ws-connector/src/WSConnector/WSConfig.hs
+++ b/libs/ws-connector/src/WSConnector/WSConfig.hs
@@ -5,15 +5,17 @@ module WSConnector.WSConfig where
 import           Prologue
 import qualified ZMQ.Bus.WS.Config as FD
 
-data Config = Config { _host     :: String
-                     , _port     :: Int
-                     , _pingTime :: Int
+data Config = Config { _host        :: String
+                     , _fromWebPort :: Int
+                     , _toWebPort   :: Int
+                     , _pingTime    :: Int
                      } deriving (Read, Show, Eq)
 
 makeLenses ''Config
 
-readWebsocketConfig config = Config host port pingTime where
-    host       = FD.host websocket
-    port       = unsafeRead (FD.port websocket) :: Int
-    pingTime   = unsafeRead (FD.pingTime websocket) :: Int
-    websocket  = FD.websocket config
+readWebsocketConfig config = Config host fromWebPort toWebPort pingTime where
+    host        = FD.host websocket
+    fromWebPort = unsafeRead (FD.fromWebPort websocket)
+    toWebPort   = unsafeRead (FD.toWebPort   websocket)
+    pingTime    = unsafeRead (FD.pingTime    websocket)
+    websocket   = FD.websocket config
--- a/libs/ws-connector/src/WSConnector/WSConfig.hs
+++ b/libs/ws-connector/src/WSConnector/WSConfig.hs
@@ -3,19 +3,20 @@
 module WSConnector.WSConfig where
 
 import           Prologue
-import qualified ZMQ.Bus.WS.Config as FD
+import qualified ZMQ.Bus.WS.Config as WSConfig
 
-data Config = Config { _host        :: String
-                     , _fromWebPort :: Int
-                     , _toWebPort   :: Int
-                     , _pingTime    :: Int
-                     } deriving (Read, Show, Eq)
+data Config = Config
+    { _host        :: String
+    , _fromWebPort :: Int
+    , _toWebPort   :: Int
+    , _pingTime    :: Int
+    } deriving (Read, Show, Eq)
 
 makeLenses ''Config
 
 readWebsocketConfig config = Config host fromWebPort toWebPort pingTime where
-    host        = FD.host websocket
-    fromWebPort = unsafeRead (FD.fromWebPort websocket)
-    toWebPort   = unsafeRead (FD.toWebPort   websocket)
-    pingTime    = unsafeRead (FD.pingTime    websocket)
-    websocket   = FD.websocket config
+    host        = WSConfig.host websocket
+    fromWebPort = unsafeRead (WSConfig.fromWebPort websocket)
+    toWebPort   = unsafeRead (WSConfig.toWebPort   websocket)
+    pingTime    = unsafeRead (WSConfig.pingTime    websocket)
+    websocket   = WSConfig.websocket config
--- a/libs/ws-connector/src/WSConnector/WSConnector.hs
+++ b/libs/ws-connector/src/WSConnector/WSConnector.hs
@@ -26,11 +26,11 @@ import qualified System.CPUTime                as CPUTime
 logger :: Logger
 logger = getLogger $moduleName
 
-timed :: IO a -> IO Double
+timed :: MonadIO m => m a -> m Double
 timed act = do
-    t0 <- CPUTime.getCPUTime
+    t0 <- liftIO $ CPUTime.getCPUTime
     act
-    t1 <- CPUTime.getCPUTime
+    t1 <- liftIO $ CPUTime.getCPUTime
     return $ fromIntegral (t1 - t0) * 1e-12
 
 handleDisconnect :: WS.ConnectionException -> IO ()
@@ -63,8 +63,8 @@ fromWebLoop conn chan = forever $ do
 fromWeb :: WS.Connection -> Unagi.InChan WSMessage -> IO ()
 fromWeb conn chan = do
     flip catch handleDisconnect $ fromWebLoop conn chan
-    let takeoverMessage = serializeFrame $ WSFrame [ControlMessage ConnectionTakeover]
-    WS.sendClose conn takeoverMessage
+    let closeMsg = serializeFrame $ WSFrame [ControlMessage ConnectionTakeover]
+    WS.sendClose conn closeMsg
 
 toWebLoop :: WS.Connection -> Unagi.OutChan WSMessage -> IO ()
 toWebLoop conn chan = forever $ do
@@ -84,7 +84,7 @@ toWeb conn chan = do
 
 sinkChan :: Unagi.OutChan a -> IO ()
 sinkChan c = forever $ do
-    !a <- Unagi.readChan c
+    !_ <- Unagi.readChan c
     return ()
 
 run :: BusEndPoints -> WSConfig.Config -> IO ()
--- a/libs/ws-connector/src/WSConnector/WSConnector.hs
+++ b/libs/ws-connector/src/WSConnector/WSConnector.hs
@@ -41,7 +41,7 @@ application clientCounter currentClient pingTime toBusChan fromBusChan pending =
     WS.forkPingThread conn pingTime
 
     let welcomeMessage = serializeFrame $ WSFrame [ControlMessage Welcome]
-    WS.sendTextData conn welcomeMessage
+    WS.sendBinaryData conn welcomeMessage
 
     fromBusListenChan <- atomically $ dupTChan fromBusChan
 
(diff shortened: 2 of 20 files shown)


libs/ws-connector/src/WSConnector/WSConfig.hs
```diff
@@ -5,15 +5,17 @@ module WSConnector.WSConfig where
 import           Prologue
 import qualified ZMQ.Bus.WS.Config as FD
```
wdanilo: The qualified name "FD" is bad for reading. I was very long thinking what it stands for and still I dont know. Reading sources is harder when using such shortucts, could you expand it please?

libs/ws-connector/src/WSConnector/WSConfig.hs
```diff
@@ -5,15 +5,17 @@ module WSConnector.WSConfig where
...
 
-data Config = Config { _host     :: String
-                     , _port     :: Int
-                     , _pingTime :: Int
+data Config = Config { _host        :: String
+                     , _fromWebPort :: Int
+                     , _toWebPort   :: Int
+                     , _pingTime    :: Int
                      } deriving (Read, Show, Eq)
```
wdanilo: Are you using strict mode here? I'm very interested how all these lazy fields affect performance.

libs/ws-connector/src/WSConnector/WSConnector.hs
```diff
@@ -21,65 +21,87 @@ import           WSConnector.Data.WSMessage    (ControlCode (..), WSMessage (..)
...
-    atomically $ mapM_ (writeTChan wsChan) $ frame ^. messages
+    mapM_ (Unagi.writeChan chan) $ frame ^. messages
 
-fromWeb :: Int -> TVar Int -> WS.Connection -> TChan WSMessage -> TChan WSMessage -> IO ()
-fromWeb clientId currentClient conn chan wsChan = do
-    flip catch handleDisconnect $ fromWebLoop clientId currentClient conn chan wsChan
+fromWeb :: WS.Connection -> Unagi.InChan WSMessage -> IO ()
+fromWeb conn chan = do
+    flip catch handleDisconnect $ fromWebLoop conn chan
+    let takeoverMessage = serializeFrame $ WSFrame [ControlMessage ConnectionTakeover]
```
wdanilo: > 80 chars

libs/ws-connector/src/WSConnector/Workers/BusWorker.hs
```diff
@@ -70,9 +71,9 @@ eitherToM = either (fail . show) return
...
 eitherToM' action = action >>= eitherToM
 
-start :: BusEndPoints -> TChan WSMessage -> TChan WSMessage -> IO ()
+start :: BusEndPoints -> Unagi.InChan WSMessage -> Unagi.OutChan WSMessage -> IO ()
 start busEndPoints fromBusChan toBusChan = do
-    exchangeIdsChan <- atomically newTChan
-    forkIO $ eitherToM' $ Bus.runBus busEndPoints $ fromBus fromBusChan exchangeIdsChan
-    forkIO $ eitherToM' $ Bus.runBus busEndPoints $ toBus   toBusChan   exchangeIdsChan
+    exchangeIdsVar <- MVar.newEmptyMVar
+    forkIO $ eitherToM' $ Bus.runBus busEndPoints $ fromBus fromBusChan exchangeIdsVar
```
wdanilo: > 80 chars

luna-studio/lib/src/WebSocket.hs
```diff
@@ -64,13 +67,19 @@ foreign import javascript safe "$1.unOnError()"
...
 
 foreign import javascript safe "$1.send($2)"
-    send :: WebSocket -> JSString -> IO ()
+    js_send :: WebSocket -> ArrayBuffer -> IO ()
+
+foreign import javascript safe "$1.connect($2, $3)"
+    connect' :: WebSocket -> JSString -> JSString -> IO ()
+
+connect :: WebSocket -> String -> String -> IO ()
+connect ws listenAddr sendAddr = connect' ws (convert listenAddr) (convert sendAddr)
```
wdanilo: > 80 chars

luna-studio/text-editor/src/TextEditor/Event/Source.hs
```diff
@@ -35,9 +36,8 @@ webSocketHandler conn = AddHandler $ \h -> do
     void $ WebSocket.onOpen conn $
         h $ Connection Connection.Opened
     void $ WebSocket.onMessage conn $ \event -> do
-        payloadJS <- WebSocket.getData event
-        let payload = fromJSString payloadJS
-            frame = BatchConnection.deserialize payload
+        payload <- WebSocket.getData event
+        let frame = BatchConnection.deserialize payload
```
wdanilo: btw, looking at all the processing here, im really curious if (and if not, why not) youre hitting performance degradation by lazy data processing. In all the code there are no fields marked strict, no computations marked strict and it seems that everything is boxed. If this is the core of web socket processing, making things strict should be our default behaviour here.

luna-studio/lib/src/WebSocket.hs
```diff
@@ -1,22 +1,25 @@
...
+                 , onOpen
+                 , send
+                 ) where
+
+import           Data.ByteString        (ByteString)
 import           Data.JSString
 import           GHCJS.Foreign.Callback
 import           GHCJS.Marshal.Pure     (PFromJSVal (..), PToJSVal (..))
 import           GHCJS.Types            (IsJSVal)
+import           GHCJS.Buffer           (toByteString, createFromArrayBuffer, fromByteString, getArrayBuffer)
```
wdanilo: > 80 chars

libs/ws-connector/src/WSConnector/Workers/BusWorker.hs
```diff
@@ -70,9 +71,9 @@ eitherToM = either (fail . show) return
...
+    forkIO $ eitherToM' $ Bus.runBus busEndPoints $ toBus   toBusChan   exchangeIdsVar
```
wdanilo: > 80 chars

libs/ws-connector/src/WSConnector/Workers/BusWorker.hs
```diff
@@ -70,9 +71,9 @@ eitherToM = either (fail . show) return
 eitherToM' :: (Monad m, Show a) => m (Either a b) -> m b
 eitherToM' action = action >>= eitherToM
 
-start :: BusEndPoints -> TChan WSMessage -> TChan WSMessage -> IO ()
+start :: BusEndPoints -> Unagi.InChan WSMessage -> Unagi.OutChan WSMessage -> IO ()
```
wdanilo: > 80 chars

libs/ws-connector/src/WSConnector/WSConnector.hs
```diff
@@ -21,65 +21,87 @@ import           WSConnector.Data.WSMessage    (ControlCode (..), WSMessage (..)
...
+    WS.sendClose conn takeoverMessage
 
-toWeb :: WS.Connection -> TChan WSMessage -> IO ()
-toWeb conn chan = flip catch handleDisconnect $ forever $ do
-    msg <- atomically $ readTChan chan
-    let webMessage = serializeFrame $ WSFrame [msg]
-    WS.sendTextData conn webMessage
+sinkChan :: Unagi.OutChan a -> IO ()
+sinkChan c = forever $ do
+    !a <- Unagi.readChan c
```
wdanilo: youre not using this variable. GHC warnings + haskell linter tell about it too. You should use `!_ <- ...` instead.

libs/ws-connector/src/WSConnector/WSConnector.hs
```diff
@@ -21,65 +21,87 @@ import           WSConnector.Data.WSMessage    (ControlCode (..), WSMessage (..)
...
-fromWeb clientId currentClient conn chan wsChan = do
-    flip catch handleDisconnect $ fromWebLoop clientId currentClient conn chan wsChan
+fromWeb :: WS.Connection -> Unagi.InChan WSMessage -> IO ()
+fromWeb conn chan = do
+    flip catch handleDisconnect $ fromWebLoop conn chan
+    let takeoverMessage = serializeFrame $ WSFrame [ControlMessage ConnectionTakeover]
+    WS.sendClose conn takeoverMessage
+
+toWebLoop :: WS.Connection -> Unagi.OutChan WSMessage -> IO ()
+toWebLoop conn chan = forever $ do
```
wdanilo: Is this a function which is only for benchmarking or for public API. If the former, it should be moved to benchamrks. In case of the later, it has poor performance characteristics – logging by concatenating Strings + not disabling logging in "production" env should not happen. If you really want to log things, you should always use loggers with data types with strict fields, no string concatenation. Such logging should always be disabled when compiling final app.

luna-studio/text-editor/src/TextEditor/Event/Source.hs
wdanilo: Moreover, in the most critical loops you should definitely use INLINE or INLINABLE pragma, so GHC could optimize it better. It would also affect worker-wrapper so more fields will be unboxed by default

tools/batch/plugins/ws-connector/src/Main.hs
```diff
@@ -1,3 +1,5 @@
+{-# LANGUAGE OverloadedStrings #-}
```
ljkania: This should be defined as default extension in `package.yaml`. You should consider doing the same with other extensions used in this pr.

libs/ws-connector/src/WSConnector/WSConfig.hs
```diff
@@ -5,15 +5,17 @@ module WSConnector.WSConfig where
 import           Prologue
 import qualified ZMQ.Bus.WS.Config as FD
 
-data Config = Config { _host     :: String
-                     , _port     :: Int
-                     , _pingTime :: Int
+data Config = Config { _host        :: String
```
ljkania: Not a huge thing (and not breaking) but we could fix layout of datatypes in touched files.

libs/ws-connector/src/WSConnector/WSConnector.hs
```diff
@@ -21,65 +21,87 @@ import           WSConnector.Data.WSMessage    (ControlCode (..), WSMessage (..)
 import qualified WSConnector.Workers.BusWorker as BusWorker
 import qualified WSConnector.WSConfig          as WSConfig
 
+import qualified System.CPUTime                as CPUTime
+
 logger :: Logger
 logger = getLogger $moduleName
 
+timed :: IO a -> IO Double
```
ljkania: Do you need to use `IO` here? Please use: `MonadIO m => ... -> m (...)` wherever possible and reasonable.

libs/ws-connector/src/WSConnector/WSConnector.hs
kustosz: Public API, logging consistent with the rest of our backend. It should not be compiled away, we do need logging in production. This is what debug levels are for. Wontfix for now, we may consider switching logging systems at some point.

libs/ws-connector/src/WSConnector/WSConfig.hs
kustosz: Global Strict for this library.

Issue #900, opened 2018-05-08
Plugging in existing visualizers


kustosz:

piotrMocz:
Is this still code review?

PR #942, opened 2018-06-11
Make focusing visualization possible #900

ljkania:
Please see: https://github.com/luna/basegl-ui/pull/4

This pr introduces the correct implementation of visualization logic. Elements are created only when necessary and updates are kept to the minimum. However there is a few issues we need to handle:
- There is some styling needed (visualization toggler is simple square, menu does not work well because of some problems with hover, background is transparent but should have color of background)
- We need to implement additional modes which show visualization full screen. To do that we need to add another layer first - this one will not be scaled and will be also used by searcher.
- We need to replace iframes with shadow dom since now focusing requires to rerender iframe loosing its state.

Awesome things this introduces:
- Extra layer for dom elements on top of base 3 layers of basegl. This one is for dom elements that should be in front (like focused visualization) and should be scaled. This layer does not affect any mouse events for base mechanism unless there is an element in this layer under the mouse.
- Two base modes for visualization: default and focused. The first one makes visualization non-responding to events and covered with basegl shape (visualization is in `layer-dom` from basegl at this state). After click on that shape visualization is focused - it is moved to top layer and events are enabled. Now everything should be under this visualization. Any outside action (but `mouseenter`, `mouseleave`, etc.) make visualization unfocus.
- Toggling node value between short value or error and visualization.

### DIFF
--- a/luna-studio/atom/lib/luna-node-editor-tab.coffee
+++ b/luna-studio/atom/lib/luna-node-editor-tab.coffee
@@ -75,6 +75,9 @@ class LunaNodeEditorTab extends View
                     tag:      evt.tag.substring(0, evt.tag.length - 5)
                     contents: evt
 
+                unless target?
+                    target = []
+
                 @nodeEditor.pushViewEvent
                     path:   path
                     target: target
--- a/luna-studio/atom/lib/luna-node-editor-tab.coffee
+++ b/luna-studio/atom/lib/luna-node-editor-tab.coffee
@@ -75,8 +75,7 @@ class LunaNodeEditorTab extends View
                     tag:      evt.tag.substring(0, evt.tag.length - 5)
                     contents: evt
 
-                unless target?
-                    target = []
+                target ?= []
 
                 @nodeEditor.pushViewEvent
                     path:   path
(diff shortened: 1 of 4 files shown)


luna-studio/atom/lib/luna-node-editor-tab.coffee
```diff
@@ -75,6 +75,9 @@ class LunaNodeEditorTab extends View
                     tag:      evt.tag.substring(0, evt.tag.length - 5)
                     contents: evt
 
+                unless target?
```
kustosz: `target ?= []`

luna-studio/node-editor/src/NodeEditor/Action/State/Visualization.hs
```diff
@@ -141,8 +141,19 @@ insertVisualization nl vis = getNodeVisualizations nl >>= \case
...
-                (not prevActive || visualizerChanged || iframeChanged)
+
+            prevModeDefault = (view mode <$> mayPrevVis) /= Just Default
+            prevModeHidden  = (view mode <$> mayPrevVis) /= Just Hidden
+            -- TODO: Remove layerChanged when iframe replaced
+            layerChanged    = case adjustedVis ^. mode of
+                Default -> (view mode <$> mayPrevVis) /= Just Default
+                Hidden  -> False
+                _       -> prevModeDefault || prevModeHidden
+            needsRegister   = isActive visEnabled errVisEnabled c adjustedVis
```
piotrMocz: I would consider a more verbose name than `c`. I see that it conflicts with the lens named `content` but I would settle for `visContent` (like `visEnabled`) or even something a little weird like `cntnt`.

PR #959, opened 2018-06-26
Scripts for building the new gui

piotrMocz:
* Added the `scripts_build/new_gui.py` file for installing the static parts of the new gui
* Added it to the `build_luna` script so that it is always called.


scripts_build/new_gui.py
```diff
@@ -0,0 +1,44 @@
...
+
+from scripts_build.atom_prepare import prep_path
+from scripts_build.common import process_context, working_directory
+
+
+new_gui_dir = prep_path('../luna-studio/basegl-ui')
+new_gui_dist = os.path.join(new_gui_dir, 'dist')
+new_gui_styles = os.path.join(new_gui_dir, 'dist/style')
+atom_package = prep_path('../dist/user-config/atom/packages/luna-studio/node_modules/node_editor_basegl')
+atom_styles = prep_path('../luna-studio/atom/styles/gen')
```
sylwiabr: why do we have to copy styles to generated styles?

scripts_build/new_gui.py
```diff
@@ -0,0 +1,44 @@
...
+import sys
+
+from scripts_build.atom_prepare import prep_path
+from scripts_build.common import process_context, working_directory
+
+
+new_gui_dir = prep_path('../luna-studio/basegl-ui')
+new_gui_dist = os.path.join(new_gui_dir, 'dist')
+new_gui_styles = os.path.join(new_gui_dir, 'dist/style')
+atom_package = prep_path('../dist/user-config/atom/packages/luna-studio/node_modules/node_editor_basegl')
```
sylwiabr: shouldn't it be added as a dependency to luna-studio?

piotrMocz: It is! I added this so that in some cases the rebuild of the whole frontend is not required (which would be the case if I let Atom manage this as its dependency). It doesn't harm us in any way when using it in the global install script but lets you use this script when you're developing `basegl-ui`'s "frontendest" part :see_no_evil:

scripts_build/new_gui.py
piotrMocz: That's because those styles are for `basegl-ui`, which is not an Atom plugin by itself. It's a dependency of an Atom plugin, namely Luna Studio. But this means that Atom won't load it's styles (only JS). That said, we could unify these styles with the old ones, from Luna Studio. But this should be done as part of purging the old GUI.

luna-studio/atom/lib/luna-node-editor-tab.coffee
```diff
@@ -208,3 +208,5 @@ class LunaNodeEditorTab extends View
             oldPath = atom.project.getPaths()[0]
             projects.temporaryProject.save (newPath) =>
                 @codeEditor.pushInternalEvent(tag: 'MoveProject', _oldPath : oldPath, _newPath: newPath)
+
+    attached: -> window.install()
```
pmlodawski: Wow, you found it! <3
I think we no longer need to assign it this to `window`

scripts_build/new_gui.py
```diff
@@ -0,0 +1,44 @@
+from distutils import dir_util
+import os
+import subprocess
+import sys
+
+from scripts_build.atom_prepare import prep_path
+from scripts_build.common import print_title, process_context, working_directory
+
+
+new_gui_dir = prep_path('../luna-studio/basegl-ui')
```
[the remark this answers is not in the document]
kustosz: This one is final, there will never be another GUI!

piotrMocz:
This is no longer relevant -- now we are building an npm package with the new gui.

PR #968, opened 2018-07-06
Dataframes visualizations

sylwiabr:


luna-studio/atom/lib/visualizers/data/base/table/table.js
```diff
@@ -68,11 +68,18 @@
     return eachHasProperLen;
   }
 
-  var genMatrix = function (data, level) {
+  var genMatrix = function (data, level, headers) {
```
piotrMocz: This is a single (singular) `header` (it comes from `data.header`) ;)

luna-studio/atom/lib/visualizers/data/base/table/table.js
```diff
@@ -68,11 +68,18 @@
...
     var result = "<tr><th></th>";
-    data[0].forEach(function (elt, ix) {
-      result += ("<th>" + ix + "</th>");
-    });
+    if (headers) {
```
piotrMocz: That, or:
```javascript
var dfHeader = headers || data[0];
dfHeader.forEach(function(elt, ix) { result += `<th>${elt}</th>`; });
```
You could even one-line it with `(headers || data[0]).forEach [...]` but I'm not so sure it's more readable.

PR #990, opened 2018-07-25
Move generate project logic to backend

pmlodawski:


libs/luna-empire/src/Empire/Commands/Graph.hs
```diff
@@ -1396,9 +1396,9 @@ readMetadata' = do
     unit     <- use Graph.clsClass
     metaRef  <- ASTRead.getMetadataRef unit
     case metaRef of
-        Just meta -> do
```
mikusp: Please refrain from fixing unrelated warnings as it hides necessary changes that are done

luna-studio/atom/lib/projects.coffee
```diff
@@ -17,8 +17,8 @@ temporaryPath         = process.env.LUNA_TMP
 tutorialsDownloadPath = process.env.LUNA_TUTORIALS
 
 temporaryProject = {
-    name: 'unsaved-luna-project'
-    path: path.join temporaryPath, 'unsaved-luna-project'
+    name: 'UnsavedLunaProject'
+    path: path.join temporaryPath, 'UnsavedLunaProject'
     srcDir: 'src'
     mainFile: 'Main.luna'
     mainContent: 'def main:\n    hello = "Hello, World!"\n    None'
```
mikusp: Is anything else than path needed here now?

PR #991, opened 2018-07-25
Use fs-no-eperm-anymore to rename files after downloading

mikusp:
Should get rid of EPERM errors while downloading tutorials on Windows


pmlodawski:
Connected issue: https://github.com/luna/luna-studio/issues/764

PR #1008, opened 2018-07-27
Prevent AddNodeResponse clean port information

pmlodawski:


luna-studio/node-editor/src/NodeEditor/Action/Basic/AddConnection.hs
```diff
@@ -34,7 +34,6 @@ localAddConnection c = do
     mayConn <- createConnectionModel src dst
     withJust mayConn $ \conn -> do
         NodeEditor.addConnection conn
-        NodeEditor.resetSuccessors $ dst ^. nodeLoc
```
mikusp: Could you explain why we don't do this anymore?

pmlodawski: Why we don't need this:
typechecker is fast enough, that nodes afected by connection will be grayed out immediately

Why am I removing it:
When you add a node, connected to another one, and typechecker update came before addNode response, this newly added node will lose its type information

PR #1021, opened 2018-08-01
Fix real sliders (hopefully the last time)

### LINKED ISSUE #994: Sliders are broken with new core
Error is:
```
Char.digitToInt: not a digit '.'
CallStack (from -prof):
while processing event
empire.graph.node.defaultValue.response
```

mikusp:
Fixes #994. Real sliders are now one of the most thoroughly tested features.

### DIFF
--- a/libs/luna-empire/src/Empire/ASTOps/Parse.hs
+++ b/libs/luna-empire/src/Empire/ASTOps/Parse.hs
@@ -23,6 +23,7 @@ import           Control.Monad.Catch          (catchAll)
 import           Data.Char                    (digitToInt)
 import qualified Data.List.Split              as Split
 import qualified Data.Text                    as Text
+import qualified Data.Scientific              as Scientific
 
 import           Empire.ASTOp                    (EmpirePass, GraphOp)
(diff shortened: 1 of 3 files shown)


libs/luna-empire/src/Empire/ASTOps/Parse.hs
```diff
@@ -202,12 +203,24 @@ parsePortDefault (Constant (TextValue s)) = do
...
+        sc          = Scientific.fromFloatDigits $ abs d
+        scString    = Scientific.formatScientific Scientific.Fixed Nothing sc
+    case Split.splitOn "." scString of
+        [int, frac] -> do
+            intPart <- Mutable.fromList $ map (fromIntegral . digitToInt) int
+            let minusLength = 1
+                dotLength   = 1
+            fracPart <- Mutable.fromList $ map (fromIntegral . digitToInt) frac
+            number   <- generalize <$> IR.number 10 intPart fracPart `withLength`
+                (length int + dotLength + length frac)
```
piotrMocz: I would extract `length int` and `length frac` into their own `let` bindings.

PR #1027, opened 2018-08-07
Fix visualization handlers

pmlodawski:


luna-studio/node-editor/src/NodeEditor/Event/View.hs
```diff
@@ -307,17 +307,12 @@ getOutPortRef = getOutPortRef' . view target where
...
 
-getVisualizerId :: ViewEvent -> VisualizerId
-getVisualizerId = getVisualizerId' . view target where
-    getVisualizerId' (Target [visId]) = convert visId
-    getVisualizerId' t = error $ "Cannot parse Target to VisualizerId (target = " <> show t <> ")"
-
 getVisualizationId :: ViewEvent -> VisualizationId
 getVisualizationId = getVisualizationId' . view target where
-    getVisualizationId' (Target [visId]) = convert visId
+    getVisualizationId' (Target (_:visId:_)) = convert visId
```
piotrMocz: The way we handle the `Target`'s field seems risky. I guess we should discuss introducing named fields. Relying on certain elements being on certain positions in a list makes the API... fragile ;)

PR #1035, opened 2018-08-10
Usable list literals

mikusp:
This patch introduces more proper list ports resolution, giving each list element one port on a node, along with working connecting/disconnecting. Connecting to already populated port will replace it, connecting to this little port at the bottom will add a new element at the end of the list, disconnecting will remove given element from the list, making the list shorter.
zrzut ekranu z 2018-08-10 13-25-59


libs/luna-empire/src/Empire/ASTOps/Builder.hs
```diff
@@ -186,12 +186,67 @@ dropBlankArgumentsAtTailPosition e beg = do
...
-    dropBlankArgumentsAtTailPosition funE beg
+    ref <- source funE
+    match ref $ \case
+        List elts -> do
+            l <- ptrListToList elts
+            let (before, after) = splitAt pos l
+                l'              = before <> drop 1 after
+            listLen <- getLayer @SpanLength ref
+            let removedArg = listToMaybe after
+            forM removedArg $ \a -> do
```
piotrMocz: I can imagine this being a separate function. If I'm correct, it would need to take 3 params: `before`, `after` and `a`. For your consideration ;)

PR #1047, opened 2018-08-16
Add phantom port key

pmlodawski:
### Pull Request Description

- This PR adds phantom ports key to gui model

### DIFF
--- a/luna-studio/node-editor/src/NodeEditor/Handler/Connect.hs
+++ b/luna-studio/node-editor/src/NodeEditor/Handler/Connect.hs
@@ -53,7 +53,7 @@ handleViewEvent evt = case evt ^. base of
         handleConnectionMouseDown def inPortRef connectionEnd
     Mouse e -> do
         let path = evt ^. View.path
-            isPort = "InPort" `elem` path || "OutPort" `elem` path
+            isPort = "InPort" `elem` path || "OutPort" `elem` path || "NewPort" `elem` path
             anyPortRef = View.getAnyPortRef evt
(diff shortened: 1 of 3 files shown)


luna-studio/node-editor/src/NodeEditor/Handler/Connect.hs
```diff
@@ -53,7 +53,7 @@ handleViewEvent evt = case evt ^. base of
         handleConnectionMouseDown def inPortRef connectionEnd
     Mouse e -> do
         let path = evt ^. View.path
-            isPort = "InPort" `elem` path || "OutPort" `elem` path
+            isPort = "InPort" `elem` path || "OutPort" `elem` path || "NewPort" `elem` path
```
ljkania: As I pointed out in another PR: why we treat "NewPort" as something different than "InPort" but "Self" is nothing special? From View (MVC) point of view they are both just special cases of "InPort".

luna-studio/node-editor/src/NodeEditor/Handler/Port.hs
```diff
@@ -24,7 +24,7 @@ handleViewEvent :: ViewEvent -> Maybe (Command State ())
 handleViewEvent evt = case evt ^. View.base of
     Mouse e -> do
         let path = evt ^. View.path
-            isPort = "InPort" `elem` path || "OutPort" `elem` path
+            isPort = "InPort" `elem` path || "OutPort" `elem` path || "NewPort" `elem` path
```
ljkania: This code is copy-paste. Please refactor to separate function

Issue #1044, opened 2018-08-14
Case for broken undo–redo


kustosz:
1. Open new project
2. Select the existing node
3. Add `reverse` node
4. Add another `reverse` node
5. Undo twice
6. With the `test` node selected, add `reverse` node.
7. Undo.
8. Redo twice.
It results in a strange state, see video in the comment. It is also worth noting that the longer you use Luna Studio and randomly undo–redo actions, you can get all sorts of strange results (including some exceptions like BreadcrumbDoesNotExist), these are however much less predictable and harder to reproduce.

kustosz:
Repro video https://drive.google.com/open?id=1kT3Wqsec9pf_A6NpnIqVdXqfffVvSCPv

piotrMocz:
I'm seeing a different (albeit still not desired, I guess) behaviour: step 8. allows me to only do one "Redo", leaving the node in a greyed-out state and preventing any further actions. If I go up to the top-level and back to `main`, it gets the type and the result. Please see the video for reference (and please pardon the notification popping up towards the end ;))

https://drive.google.com/file/d/1Tj23w3aYHpPXKftM_v5Cqh0Yy15MNcvf/view?usp=sharing

I'm moving this task back to `In Progress` -- maybe I'm wrong and maybe that's a different task altogether. @mikusp feel free to close it and open another one, if needed.

mikusp:
Redoing only once is correct. However, the node should be typechecked so I'll move that to backlog and see what's going on.

mikusp:
We can't reproduce this issue

PR #1057, opened 2018-08-22
Clean redo stack on new action #1044

mikusp:
### Pull Request Description

With previous implementation, some stale actions could end up in redo stack longer than expected. When using redo afterwards, these action were executed in a state when they could no longer make sense. To prevent that, whenever a new action is performed by a user, redo stack is cleaned for that user.


libs/undo-redo/src/Handlers.hs
```diff
@@ -48,8 +48,14 @@ import qualified LunaStudio.Data.Graph                   as Graph
 import qualified LunaStudio.Data.Node                    as Node
 import           LunaStudio.Data.Port                    (OutPortIndex (Projection))
 import           LunaStudio.Data.PortRef                 (AnyPortRef (InPortRef'), OutPortRef (..))
+import qualified System.Log.MLogger                      as Logger
 import           Prologue                                hiding (throwM)
 
+
+logger :: Logger.Logger
```
piotrMocz: Are we logging something here?

ljkania:
PR is merged. Should this issue be moved?

PR #1060, opened 2018-08-27
Enable connections to subports

pmlodawski:
This PR enable displaying connections to subports. It also fixes related crash

### DIFF
--- a/luna-studio/node-editor-view/src/NodeEditor/View/Key.hs
+++ b/luna-studio/node-editor-view/src/NodeEditor/View/Key.hs
@@ -61,3 +61,11 @@ instance Convertible Key AnyPortId where
 
 instance Convertible UUID String where
     convert = show
+
+sliceAnyPortId :: AnyPortId -> AnyPortId
+sliceAnyPortId (InPortId' portId) = InPortId' $ slicePortId portId
+sliceAnyPortId (OutPortId' portId) = OutPortId' $ slicePortId portId
+
+slicePortId :: [a] -> [a]
+slicePortId [] = []
+slicePortId (h:_) = [h]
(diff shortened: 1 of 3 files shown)


luna-studio/node-editor-view/src/NodeEditor/View/Key.hs
```diff
@@ -60,4 +60,12 @@ instance Convertible Key AnyPortId where
...
 instance Convertible UUID String where
-    convert = show
\ No newline at end of file
+    convert = show
+
+sliceAnyPortId :: AnyPortId -> AnyPortId
+sliceAnyPortId (InPortId' portId) = InPortId' $ slicePortId portId
+sliceAnyPortId (OutPortId' portId) = OutPortId' $ slicePortId portId
+
+slicePortId :: [a] -> [a]
```
ljkania: This name is wrong. `slice` function does something different. Consider changing name to `headPortId` and similar changes in the rest of changes.

ljkania: Also it works for any type. Don't we have anything like `safeHead` to use here?

luna-studio/node-editor-view/src/NodeEditor/View/Key.hs
```diff
@@ -60,4 +60,12 @@ instance Convertible Key AnyPortId where
     convert (Key key) = error $ "Cannot parse " <> key <> " to AnyPortId"
 
 instance Convertible UUID String where
-    convert = show
\ No newline at end of file
+    convert = show
+
+sliceAnyPortId :: AnyPortId -> AnyPortId
```
ljkania: Please move this function to `LunaStudio.Data.Port` as getter to `AnyPortId`. Also consider rename.

PR #1074, opened 2018-09-18
Dataframes support in Luna Studio

sylwiabr:
### Pull Request Description

This PR is adding the support for Dataframes in Luna Studio

### Important Notes

- Support the viisualizations for Dataframes
- Bump luna and StdLib with Lquery
- Fix the compilation on Windows

### Checklist

- [ x ] The documentation has been updated if necessary.
- [ x ] The code conforms to the [Luna Style Guide](https://github.com/luna/wiki/blob/master/code-style/01.general.md) if it is Haskell.
- [ x ] The code has been tested where possible.


env/Std/src/Base.luna
```diff
@@ -1047,7 +1053,15 @@ class JSON:
     def toJSON: self
 
     # Dumps a `JSON` structure into a `Text` object.
-    def toText: renderJSON self
+    def toText: case self of
+        JSONNumber d: d.toText
+        JSONString s: s.escapeJSON
+        JSONArray  l: "[" + (l . map (x : x.toText) . intersperse "," . fold "" (+)) + "]"
```
ljkania: Why don't we use `concat` instead of `fold "" (+)`?

ljkania: Also, AFAIR: `map .toText` should work here too

env/Std/src/Base.luna
```diff
@@ -1047,7 +1053,15 @@ class JSON:
...
+        JSONBool   b: case b of
+            True:  "true"
+            False: "false"
+        JSONNull: "null"
+        JSONObject m: "{" + (m . toList . map ((Tuple2 k v): k.escapeJSON + ": " + v.toText) . intersperse "," . fold "" (+)) + "}"
```
ljkania: ditto

luna-studio/atom/lib/visualizers/data/base/config.js
```diff
@@ -1,5 +1,5 @@
 module.exports = function (type) {
-    var listVis  = (type.constructor === "List" || type.constructor === "Dataframe") ? [{name: "table", path: "table/table.html"}] : [];
+    var listVis  = (type.constructor === "List" || type.constructor === "Table" || type.constructor === "Column") ? [{name: "table", path: "table/table.html"}] : [];
```
ljkania: Consider making a list of accepted constructors and using `contains` (or equivalent - I don't remember exact function name). There is a lot of repetition.

luna-studio/atom/lib/visualizers/data/base/table/table.js
```diff
@@ -80,7 +80,18 @@
...
     };
     result += "</tr>";
-    data.forEach(function (row, ix) {
+    table = []
+
+    data.forEach(function(d, i) {
+      d.forEach(function(elem, idx) {
+        if (!table[idx]) {
+          table[idx] = []
+        }
```
ljkania: `table[idx] = table[idx] || []`

env/Std/src/Base.luna
sylwiabr: We should consider this when we will be making changes in stdlib for time series in luna

env/Std/src/Base.luna
sylwiabr: like above

luna-studio/atom/lib/visualizers/data/base/config.js
```diff
@@ -1,5 +1,5 @@
 module.exports = function (type) {
-    var listVis  = (type.constructor === "List" || type.constructor === "Table" || type.constructor === "Column") ? [{name: "table", path: "table/table.html"}] : [];
+    var listVis  = (["List", "Table", "Column"].includes(type.constructor)) ? [{name: "table", path: "table/table.html"}] : [];
```
ljkania: Love it <3

PR #1102, opened 2018-09-22
Initial searcher improvement

ljkania:
### Pull Request Description

This PR is first part of series of improvements to searcher. Searcher now is always informed about imports changes therefore can propose hints better (based on information if given module is imported). Also with [this commit](https://github.com/luna/fuzzy-text/commit/1fe94d4d8b106b2ab37c65de9200b3f0404ac2ca) it develops sorting of hints so perfect match can always appear first. This is a temporary solution and it's meant to be replaced with proper scoring.

This PR should be considered as a solution to: [bug about not updated info about imports](https://github.com/luna/luna-studio/issues/1086) and [alternative proposal to exact match always being displayed first](https://github.com/luna/luna-studio/issues/1075).

The main values of this PR are:
- searcher is immediately updated when library is imported so we can score their contents higher right away
- when there is a hint matching exactly users input it will be positioned as first hint if it comes from imported module. This way we invite users to use already imported libraries (which they usually want to do) but allowing them to see and select other hint which will result in new import. With previous improvement this hint will be much higher scored in next search.

### Important Notes

- The minimal representation of graph now also contains info about available imports
- LunaStudio.Data.Graph now has a new field called `imports`. This class is very basic so it is important change but shouldn't affect user in any way (this is internal luna-studio class).

### Checklist

### DIFF
--- a/libs/batch/plugins/luna-empire/src/Empire/Server/Server.hs
+++ b/libs/batch/plugins/luna-empire/src/Empire/Server/Server.hs
@@ -155,14 +155,11 @@ catchAllExceptions act = try act
 withDefaultResult' :: (GraphLocation -> Empire Graph) -> GraphLocation
     -> Empire a -> Empire Diff
 withDefaultResult' getFinalGraph location action = do
-    let addImports imps g = g & GraphAPI.imports .~ imps
-    oldImports <- Graph.getAvailableImports location
-    oldGraph   <- (_Left %~ Graph.prepareGraphError) <$> catchAllExceptions
-        (addImports oldImports <$> Graph.getGraphNoTC location)
+    oldGraph <- (_Left %~ Graph.prepareGraphError)
+        <$> catchAllExceptions (Graph.getGraphNoTC location)
     void action
-    newImports <- Graph.getAvailableImports location
-    newGraph   <- (_Left %~ Graph.prepareGraphError) <$> catchAllExceptions
-        (addImports newImports <$> getFinalGraph location)
+    newGraph <- (_Left %~ Graph.prepareGraphError)
+        <$> catchAllExceptions (getFinalGraph location)
     pure $ diff oldGraph newGraph
 
 withDefaultResult :: GraphLocation -> Empire a -> Empire Diff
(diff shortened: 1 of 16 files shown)


libs/luna-empire/test/FileModuleSpec.hs
```diff
@@ -919,7 +919,7 @@ spec = around withChannels $ parallel $ do
                         4
                     |]
             in specifyCodeChange initialCode initialCode $ \loc -> do
-                imports <- Graph.getAvailableImports loc
+                imports <- Graph.getAvailableImports $ loc
```
mikusp: Unnecessary $

libs/luna-empire/src/Empire/Commands/GraphBuilder.hs
```diff
@@ -82,19 +82,21 @@ buildGraph = do
...
-    API.Graph
+    monads      <- buildMonads
+    pure $ API.Graph
         nodes
         (uncurry API.Connection <$> connections)
         (Just inE)
         (Just outE)
-        <$> buildMonads
+        monads
+        mempty
```
mikusp: You should populate this field with real imports

libs/batch/plugins/luna-empire/src/Empire/Server/Server.hs
```diff
@@ -155,11 +155,14 @@ catchAllExceptions act = try act
...
+    let addImports imps g = g & GraphAPI.imports .~ imps
+    oldImports <- Graph.getAvailableImports location
+    oldGraph   <- (_Left %~ Graph.prepareGraphError) <$> catchAllExceptions
+        (addImports oldImports <$> Graph.getGraphNoTC location)
     void action
-    newGraph <- (_Left %~ Graph.prepareGraphError)
-        <$> catchAllExceptions (getFinalGraph location)
+    newImports <- Graph.getAvailableImports location
+    newGraph   <- (_Left %~ Graph.prepareGraphError) <$> catchAllExceptions
+        (addImports newImports <$> getFinalGraph location)
```
mikusp: It would be better to include imports in `getGraph` results

libs/luna-empire/test/FileModuleSpec.hs
ljkania: Fixed

libs/luna-empire/src/Empire/Commands/GraphBuilder.hs
ljkania: This is hard to do it here, but I've developed `getGraph` function so it always fill imports after getting graph.

ljkania:
@mikusp: please take a look again.

libs/batch/plugins/luna-empire/src/Empire/Server/Server.hs
ljkania: Done.

PR #1134, opened 2018-09-24
Fix reading imports from a file #1101

### LINKED ISSUE #1101: Importing modules below initial Std.Base will result in non exhaustive pattern match error
### General Summary

This bug basically messes up with the parser and the visualization part in general, as soon as this bug happens, you will have to start a new project in order to work on the code, even deleting the whole file content won't help.

### Steps to Reproduce

Please list the reproduction steps for your bug. For example:

1. Create a new project.
2. import a module **BELOW** `import Std.Base` 
3. Observe

### Expected Result

This error shouldn't raise at all!

### Actual Result
image

In this picture I tried to delete the file content and start over after reproducing the bag

### Luna Version
Luna studio v1.4.101.6

mikusp:
### Pull Request Description

Fixes #1101, as now it is a possibility that Invalid node is present in imports.

### Checklist

### DIFF
--- a/libs/luna-empire/src/Empire/Commands/Graph.hs
+++ b/libs/luna-empire/src/Empire/Commands/Graph.hs
@@ -1688,7 +1688,7 @@ nativeModuleName = "Native"
 getImportsInFile :: ClassOp [Text]
 getImportsInFile = do
     unit <- use Graph.clsClass
-    matchExpr unit $ \case
+    imports <- matchExpr unit $ \case
         Unit imps _ _ -> do
             imps' <- source imps
             matchExpr imps' $ \case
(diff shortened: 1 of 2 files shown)


libs/luna-empire/src/Empire/Commands/Graph.hs
```diff
@@ -1703,7 +1703,10 @@ getImportsInFile = do
                                     ImportSrc n -> case n of
                                         Term.Absolute n -> do
                                             let n' = convert n
-                                            return $ convert $ nameToString n'
+                                            return $ Just $ convert $ nameToString n'
+                                    _ -> return Nothing
+                            _ -> return Nothing
+    return $ catMaybes imports
```
ljkania: This code is ugly and should be refactored. It actually is refactored in another PR.

Consider this:
```
getImportsInFile :: ClassOp [Text]
getImportsInFile = matchUnit =<< use Graph.clsClass where
    matchUnit unit = matchExpr unit $ \(Unit imps _ _) -> 
        matchImports =<< source imps
    matchImports imps = matchExpr imps $ \(ImportHub imps') -> 
        mapM matchImport =<< mapM source =<< ptrListToList imps'
    matchImport imp = matchExpr imp $ \case
        Import absolute _ -> matchAbsolute =<< source absolute
        _                 -> pure Nothing
    matchAbsolute a = matchExpr a $ \case
        ImportSrc (Term.Absolute n) -> pure . Just . nameToText $ convert n
        _                           -> pure . Nothing

```

mikusp: Merge the other PR and I'll resolve the conflict

ljkania: for my own sake I need to write this down, you could do: `fmap catMaybes . matchUnit =<< use Graph.clsClass where`

Issue #1132, opened 2018-09-24
Working with node graph completely messes the text


wdanilo:
### Steps to Reproduce

1. New project
2. top level -> tab -> `def foo`
3. get inside -> tab -> `15`
4. navigate to main and remove the hello wrold node 

### Expected Result

Main should be empty

### Actual Result

The code we get:

```haskell
import Std.Base

def foo:
    number1 = 15
    None

def  me!"
    None

```

### Luna Version
1.4.101.6

mikusp:
It's the same issue as in #1131, closing as duplicate.

PR #1135, opened 2018-09-24
Fix length of function after validating it #1131

### LINKED ISSUE #1131: Creating a toplevel function breakes the code
### Steps to Reproduce

1. New project
2. Navigate to top level
3. tab -> `def foo` 
4. get inside
5. tab -> 15
6. go to top level

### Expected Result

You should see two nodes, `main` and `foo`

### Actual Result

There are 2 nodes `foo` and `e <<`

### Luna Version
1.4.101.6

mikusp:
### Pull Request Description

Fixes #1131 and #1132 

### Checklist

### DIFF
--- a/libs/luna-empire/src/Empire/ASTOps/Builder.hs
+++ b/libs/luna-empire/src/Empire/ASTOps/Builder.hs
@@ -360,6 +360,7 @@ ensureFunctionIsValid = do
                 matchExpr self $ \case
                     ASGFunction _ _ b -> putLayer @SpanOffset b $
                         fromIntegral $ length (":\n    " :: String)
+                Code.gossipLengthsChangedBy (fromIntegral $ Text.length section) =<< target b
         _ -> return ()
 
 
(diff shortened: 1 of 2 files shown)


libs/luna-empire/src/Empire/ASTOps/Builder.hs
```diff
@@ -360,6 +360,7 @@ ensureFunctionIsValid = do
                 matchExpr self $ \case
                     ASGFunction _ _ b -> putLayer @SpanOffset b $
                         fromIntegral $ length (":\n    " :: String)
+                Code.gossipLengthsChangedBy (fromIntegral $ Text.length section) =<< target b
```
piotrMocz: What's this type here? If that's an `Integer`, I would be inclined to change it into a simple `Int`, for both performance and code readability.

mikusp: It's Delta, newtype for Word64

Issue #1138, opened 2018-09-25
Show arguments names on expanded nodes


sylwiabr:
### General Summary

When we are working on expanded nodes we can see our arguments as `arg0`, `arg1` etc. We want to know what we have to connect/set as an argument so arguments names instead will be useful.

### Motivation

Easy use of nodes with missing arguments/expanded nodes

piotrMocz:
This works only in one case that I'm aware of: fully applied, user-defined, top-level function, all entered as a single expression over the node.

## Cases that break:

### User-defined top-level function

#### Unapplied function
1. Start with an empty project.
2. Definine a function in the text representation, for example:
```python
def foo x y:
    x + y
```
3. Create, in the graph editor, a node with just this function (unapplied):
    3.1. Hit "Tab" and write `foo`.
    3.2. Hit "Enter" to accept the expression.
    3.3. Hit "Enter" once again to expand the node -- you'll see that the arguments are named `arg0` and `arg1` and not `x` and `y`.

#### Connecting a node in place of an argument of a fully applied function

1. Repeat steps 1 and 2 from above
2. Create a fully applied function:
    2.1. Hit "Tab" and write `foo 12 34`
    2.2. Hit "Enter"
    2.3. Expand the node by hitting "Enter" (you'll notice that the argument names are displayed correctly).
3. Create a node with just a number (hit "Tab", enter `123` and hit "Enter").
4. Click and drag the number node's border to connect it to one of the ports of the function node.
5. Notice how the argument names changed to `arg0` and `arg1`.

#### Modifying a fully applied function using the expanded node controls

1. Repeat all of the steps 1 and 2 from above.
2. In the expanded node, move one of the sliders to change the value.
3. Notice how the argument names switch to `arg0` and `arg1`.

### Functions from the Standard Library

#### Partially applied

1. Hit "Tab" and select one of the top-level functions from the standard library (not methods). For example, we could use `readFile`, which has one argument: `f`.
2. Hit "Enter" to select it without any arguments.
3. Hit "Enter" once again to expand the node: the argument name is `arg0`.

#### Fully applied

Same as above, but enter an argument to the function (a path to any file on your disk).
<img width="251" alt="screen shot 2018-09-28 at 12 43 31" src="https://user-images.githubusercontent.com/3109645/46204064-2611c700-c31c-11e8-9767-f5f9d50d8f66.png">

@mikusp if I'm somehow using this wrong, please correct me. If not, please move the task to the "In Progress" column.

[...]

PR #1143, opened 2018-09-25
Show real names of arguments in application #1138

mikusp:
### Pull Request Description

This pull request sends proper arguments names in expanded function application nodes.

### Important Notes

I'm not a fan of threading this TCFunResolver through a few functions so I'm open to suggestions here.

### Checklist

### DIFF
--- a/libs/batch/plugins/luna-empire/src/Empire/Server.hs
+++ b/libs/batch/plugins/luna-empire/src/Empire/Server.hs
@@ -145,7 +145,7 @@ startTCWorker :: Empire.CommunicationEnv -> Bus ()
 startTCWorker env = liftIO $ do
     let reqs = env ^. Empire.typecheckChan
     pmState <- Graph.defaultPMState
-    let interpreterEnv = Empire.InterpreterEnv (return ()) (error "startTCWorker: clsGraph") [] def def def
+    let interpreterEnv = Empire.InterpreterEnv (return ()) (error "startTCWorker: clsGraph") [] def def def def
         commandState   = Graph.CommandState pmState interpreterEnv
     void $ Empire.evalEmpire env commandState $ do
         Typecheck.makePrimStdIfMissing
--- a/libs/batch/plugins/luna-empire/src/Empire/Server.hs
+++ b/libs/batch/plugins/luna-empire/src/Empire/Server.hs
@@ -145,7 +145,14 @@ startTCWorker :: Empire.CommunicationEnv -> Bus ()
 startTCWorker env = liftIO $ do
     let reqs = env ^. Empire.typecheckChan
     pmState <- Graph.defaultPMState
-    let interpreterEnv = Empire.InterpreterEnv (return ()) (error "startTCWorker: clsGraph") [] def def def def
+    let interpreterEnv = Empire.InterpreterEnv
+                            (pure ())
+                            (error "startTCWorker: clsGraph")
+                            mempty
+                            def
+                            def
+                            def
+                            def
         commandState   = Graph.CommandState pmState interpreterEnv
     void $ Empire.evalEmpire env commandState $ do
         Typecheck.makePrimStdIfMissing
--- a/libs/luna-empire/src/Empire/Commands/GraphBuilder.hs
+++ b/libs/luna-empire/src/Empire/Commands/GraphBuilder.hs
@@ -346,23 +346,22 @@ extractAppArgNames :: NodeRef -> Maybe TCFunResolver -> GraphOp [Maybe String]
 extractAppArgNames node funResolve = go [] node
     where
         go :: [Maybe String] -> NodeRef -> GraphOp [Maybe String]
-        go vars node = do
-            match node $ \case
-                ResolvedDef mod n -> do
-                    let fun = case funResolve of
-                            Just f -> f mod n
-                            _      -> Nothing
-                    case fun of
-                        Just f -> extractArgNames (generalize f) funResolve
-                        _      -> pure []
-                App f a -> do
-                    varName <- safeGetVarName =<< source a
-                    go (varName : vars) =<< source f
-                Lam{}   -> extractArgNames node funResolve
-                Cons{}  -> pure vars
-                Var{}   -> pure vars
-                Acc{}   -> pure vars
-                _       -> pure []
+        go vars node = match node $ \case
+            ResolvedDef mod n -> do
+                let fun = case funResolve of
+                        Just f -> f mod n
+                        _      -> Nothing
+                case fun of
+                    Just f -> extractArgNames (generalize f) funResolve
+                    _      -> pure []
+            App f a -> do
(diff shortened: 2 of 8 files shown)


libs/luna-empire/src/Empire/Commands/GraphBuilder.hs
```diff
@@ -316,18 +321,18 @@ safeGetVarName node = do
         (\(e :: ASTRead.NoNameException) -> pure Nothing)
     pure name
 
-extractArgNames :: NodeRef -> GraphOp [Maybe String]
-extractArgNames node = do
+extractArgNames :: NodeRef -> Maybe TCFunResolver -> GraphOp [Maybe String]
```
piotrMocz: Yeah, this is why optional/keyword arguments are a cool thing ;D

libs/luna-empire/src/Empire/Commands/GraphBuilder.hs
```diff
@@ -337,26 +342,34 @@ extractArgNames node = do
...
-        go vars node = match node $ \case
-            App f a -> do
-                varName <- safeGetVarName =<< source a
-                go (varName : vars) =<< source f
-            Lam{}   -> extractArgNames node
-            Cons{}  -> pure vars
-            Var{}   -> pure vars
-            Acc{}   -> pure vars
-            _       -> pure []
+        go vars node = do
```
piotrMocz: This do is not necessary, I'm all for reducing indentation levels ;)

libs/batch/plugins/luna-empire/src/Empire/Server.hs
```diff
@@ -145,7 +145,7 @@ startTCWorker :: Empire.CommunicationEnv -> Bus ()
 startTCWorker env = liftIO $ do
     let reqs = env ^. Empire.typecheckChan
     pmState <- Graph.defaultPMState
-    let interpreterEnv = Empire.InterpreterEnv (return ()) (error "startTCWorker: clsGraph") [] def def def
+    let interpreterEnv = Empire.InterpreterEnv (return ()) (error "startTCWorker: clsGraph") [] def def def def
```
ljkania: use `pure` instead of `return`

ljkania: also `mempty` > `[]`, and you may want to use `mempty` whenever `def == mempty`

libs/luna-empire/src/Empire/Commands/GraphBuilder.hs
```diff
@@ -337,15 +342,22 @@ extractArgNames node = do
...
-extractAppArgNames node = go [] node
+extractAppArgNames :: NodeRef -> Maybe TCFunResolver -> GraphOp [Maybe String]
+extractAppArgNames node funResolve = go [] node
     where
         go :: [Maybe String] -> NodeRef -> GraphOp [Maybe String]
         go vars node = match node $ \case
+            ResolvedDef mod n -> do
+                let fun = case funResolve of
+                        Just f -> f mod n
+                        _      -> Nothing
```
ljkania: `let fun = (\f -> f mod n) <$> funResolve`

libs/luna-empire/src/Empire/Commands/Typecheck.hs
```diff
@@ -146,13 +147,15 @@ runInterpreter path imports = do
...
+        units <- use $ Graph.userState . mappedUnits
+        let resolveFun mod n = units ^? ix mod . to Unit._definitions . wrapped . ix n . Def.documented . Def._Body
         zoomCommand clsGraph $ withRootedFunction uuid $ runInternalBreadcrumb (Breadcrumb rest) $ do
             (inEdge, outEdge) <- use $ Graph.userState . Graph.breadcrumbHierarchy . BH.portMapping
             (updates, errors) <- runASTOp $ do
                 sidebarUpdates <- (\x y -> [x, y]) <$> GraphBuilder.buildInputSidebarTypecheckUpdate  inEdge
                                                    <*> GraphBuilder.buildOutputSidebarTypecheckUpdate outEdge
                 allNodeIds  <- uses Graph.breadcrumbHierarchy topLevelIDs
-                nodeUpdates <- mapM GraphBuilder.buildNodeTypecheckUpdate allNodeIds
+                nodeUpdates <- mapM (GraphBuilder.buildNodeTypecheckUpdate resolveFun) allNodeIds
```
ljkania: You can get rid of braces using `forM`

libs/batch/plugins/luna-empire/src/Empire/Server.hs
wdanilo: According to @ljkania comments:
- `pure` instead of `return` because `pure` has weaker constraints of `Applicative` while `return` requires `Monad`. It's not a big deal but as these functions do exactly the same, always no matter what, we agreed to use one instead of both. That's why `return` has a warning in Prologue. We've been chatting a lot aobut it on discord in the past, but if you missed the conversation, I'd be happy to chat again.

- `mempty` is better than `[]` because again its more general. If you start using something other than lists here the code will still work. While writing code using such polymorphic functions it makes it easier to refactor and maintain.

wdanilo: Can we please just remove `error` from our code? Even if we just touch the line - `error` should never be used.

libs/luna-empire/src/Empire/Commands/Typecheck.hs
```diff
@@ -146,13 +147,15 @@ runInterpreter path imports = do
 updateNodes :: GraphLocation -> Command InterpreterEnv ()
 updateNodes loc@(GraphLocation _ br) = case br of
     Breadcrumb (Definition uuid:rest) -> do
+        units <- use $ Graph.userState . mappedUnits
+        let resolveFun mod n = units ^? ix mod . to Unit._definitions . wrapped . ix n . Def.documented . Def._Body
```
wdanilo: This is very hard to understand. I'd suggest splitting it to few variables.

libs/batch/plugins/luna-empire/src/Empire/Server.hs
mikusp: With what? There's no way to create clsGraph here.

mikusp:
TCFunResolver is passed whenever we run this function from typechecker thread, as it's the only situation when we have resolved the function to concrete def. Let's say we have a code like
```
def foo bar baz:
    bar + baz

def main:
    a = foo 1 2
    None
```
We know that `foo` in `foo 1 2` resolves to `CurrentProject.foo` only after typechecking. Then, when typechecker computes new ports for a node, it has access to information about function being called, and only then it can compute proper argument names.

libs/luna-empire/src/Empire/Commands/Typecheck.hs
mikusp: Will do.

Issue #1096, opened 2018-09-21
Missing connection to section field


wdanilo:
### General Summary

Write the following code using text editor:

```haskell
import Std.Base

def main:
    foo = 1 
    foo +
```

The result is:

![9](https://user-images.githubusercontent.com/1623053/45886763-37a61c80-bdba-11e8-8674-c55b57175086.png)

It's critical cause it's very simple expression and the results are totally misleading.

### Expected Result

- The nodes refer to the same variable and should be connected.

### Luna Version

1.4

piotrMocz:
Works like a charm ;)

PR #1151, opened 2018-09-28
Show connections to Sections #1096

mikusp:
### Pull Request Description

Shows connections to expressions that are {Left,Right}Section terms.

### Checklist

### DIFF
--- a/libs/luna-empire/src/Empire/ASTOps/Deconstruct.hs
+++ b/libs/luna-empire/src/Empire/ASTOps/Deconstruct.hs
@@ -71,8 +71,8 @@ extractArguments' FApp expr = match expr $ \case
         return $ arg' : args
     Grouped g -> source g >>= extractArguments' FApp
     Acc t n -> source n >>= extractArguments' FApp
-    RightSection _op n -> source n >>= \a -> return [a]
-    LeftSection _op n  -> source n >>= \a -> return [a]
+    RightSection _op n -> source n >>= \a -> pure [a]
+    LeftSection  _op n -> source n >>= \a -> pure [a]
     _       -> return []
 extractArguments' FLam expr = match expr $ \case
     Lam b a -> do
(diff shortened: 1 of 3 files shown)


libs/luna-empire/src/Empire/ASTOps/Deconstruct.hs
```diff
@@ -71,6 +71,8 @@ extractArguments' FApp expr = match expr $ \case
         return $ arg' : args
     Grouped g -> source g >>= extractArguments' FApp
     Acc t n -> source n >>= extractArguments' FApp
+    RightSection _op n -> source n >>= \a -> return [a]
+    LeftSection _op n  -> source n >>= \a -> return [a]
```
ljkania: Please fix indent of args so they are aligned. Also this is duplicated code. We should refactor it.

ljkania: Use `pure` instead of `return`

libs/luna-empire/src/Empire/Commands/GraphBuilder.hs
```diff
@@ -622,7 +622,10 @@ deepResolveInputs nid ref portRef@(InPortRef loc id) = do
             <$> (filter ((/= nid) . view srcNodeId) currentPortResolution)
         unfilteredPortConn = (, portRef) <$> currentPortResolution
     args       <- ASTDeconstruct.extractAppPorts ref
-    argsConns  <- forM (zip args [0..]) $ \(arg, i)
+    startingPortNumber <- match ref $ \case
+        LeftSection{} -> return 1
+        _             -> return 0
```
ljkania: Use `pure`.

libs/luna-empire/src/Empire/Commands/GraphBuilder.hs
```diff
@@ -622,7 +622,10 @@ deepResolveInputs nid ref portRef@(InPortRef loc id) = do
...
+    argsConns  <- forM (zip args [startingPortNumber..]) $ \(arg, i)
```
ljkania: because layout of functions changed indentation before `<-` seems strange now in this section

PR #1165, opened 2018-10-03
Fix incorrect ports for pattern match on custom datatype

### LINKED ISSUE #1100: Functions being displayed as nodes will get their names displayed without the last chars (Not constant)
Watch as 3 functions are missing last character in the visual part vs the textual part
image

EDIT: it looks like the only first function is the one to not being affected.

### LINKED ISSUE #1110: Tuple visualization is weird when pattern matching, and has a random behavior
### General Summary

Node which visual a Tuple will have a random behavior, sometime the node will be there, sometime it won't, sometime it will show it's strings, sometimes it won't

### Steps to Reproduce

```elm
def returnTuple:
  ("a", "b")

def main:
  (a, b) = returnTuple # observe the visualization of this
```

### Expected Result

The node will behave just like other nodes

### Actual Result
image

### Luna Version

- Luna Studio 1.4.101.4

ljkania:
### Pull Request Description

This pull request fixes #1110 which turned out to be a problem with handling response from typechecker. Because I needed to introduce tests for this issue I've also provided with wrapper function for all preparations and execution that needs to be done to get results from TC.

### Important Notes

- Pattern match for custom data types works and shows correct ports.
- An easy to use testing function which fires TC is now available so testing should be easier.

### Checklist

### DIFF
--- a/libs/luna-empire/test/Test/Hspec/Empire.hs
+++ b/libs/luna-empire/test/Test/Hspec/Empire.hs
@@ -35,7 +35,6 @@ import Empire.Empire                   (InterpreterEnv (InterpreterEnv))
 import Luna.Package.Structure.Generate (genPackageStructure)
 import LunaStudio.Data.GraphLocation   (GraphLocation (GraphLocation))
 import System.Directory                (canonicalizePath, getCurrentDirectory)
-import System.Environment              (lookupEnv, setEnv)
 import System.FilePath                 (takeDirectory, (</>))
 import Test.Hspec                      (Expectation, Spec, SpecWith, around,
                                         describe, parallel, shouldBe)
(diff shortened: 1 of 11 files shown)


libs/luna-empire/test/Test/Hspec/Empire.hs
```diff
@@ -87,22 +102,81 @@ testCaseWithMarkers
...
         codeCheck expectedCode resultCode
+        pmState <- Graph.defaultPMState
+        let commandState = CommandState pmState
+                $ InterpreterEnv (pure ()) clsGraph mempty def def def def
+        updatedState <- execEmpire env commandState $
+            Typecheck.runNoCleanUp gl clsGraph rooted False False
+        let updatedClsGraph
+                = updatedState ^. Graph.userState . Empire.clsGraph
+        void . evalEmpire env state $ do
+            libPath <- fmap (view Library.path . head) listLibraries
```
mikusp: On the second thought, I'm pretty sure you can use filePath from `gl` instead of using head on libraries opened.

PR #1178, opened 2018-10-04
Refactor substitute

ljkania:
### Pull Request Description

- This PR starts refactor on Empire.Commands.Graph. Reason for it is my beginning of work with markers. This is minimal part of Empire.Commands.Graph that needed to be refactored in order to separate code substitute functions. All other PRs should have higher priorities than this one. This PR can become a fix to marker problem if hangs here for longer time and I am able to implement it before merge.

### Important Notes

- Code cleaning. 

### Checklist


libs/luna-studio-common/src/LunaStudio/Data/GraphLocation.hs
```diff
@@ -20,3 +24,44 @@ instance FromJSON GraphLocation where parseJSON = Lens.parse
...
+(|>) :: GraphLocation -> BreadcrumbItem -> GraphLocation
+(|>) = appendBreadcrumbItem
+
+appendBreadcrumbItem :: GraphLocation -> BreadcrumbItem -> GraphLocation
+appendBreadcrumbItem (GraphLocation file bc) item
+    = GraphLocation file . coerce . (<> [item]) $ coerce bc
+
+infixl 5 |>*
+(|>*) :: GraphLocation -> NodeId -> GraphLocation
+(|>*) = appendLambda
```
mikusp: I don't really like how this operator looks, and given that (|>\*) is used a whole lot more compared to (|>), please consider changing (|>\*) back to (|>) and renaming (|>) to something else, maybe (|>|)?

libs/luna-studio-common/src/LunaStudio/Data/Breadcrumb.hs
```diff
@@ -25,7 +28,7 @@ data Named a = Named
 
 newtype Breadcrumb a = Breadcrumb
     { _items :: [a]
-    } deriving (Eq, Generic, Ord, Show)
+    } deriving (Eq, Foldable, Generic, Ord, Show)
```
mikusp: Is this required somewhere?

ljkania: Yeah, with this you can easily ask for `null bc`

libs/luna-studio-common/src/LunaStudio/Data/GraphLocation.hs
ljkania: Yeah, but `|>` is shorter and since one of our function takes `BreadcrumbItem` and rest of them arguments that are used to create new `BreadcrumbItem` it seems more natural to use shorter one for `BreadcrumbItem` append.

Issue #1093, opened 2018-09-21
Creating empty function on nodes fail


wdanilo:
### General Summary

I'm trying to create a function in `main` using node editor. It's a critical bug because it makes creating functions impossible.

### Steps to Reproduce

1. Open new project
2. Tab -> `def foo`. Here is the first bug - it should already create an empty function, not raise an error.
3. Rename `def foo` to `def foo:` 
4. Double click on the node
5. Tab -> `None` and connect it to output. Here is another error - every expression we've created is in `main`, not in `foo`.
6. Using text editor we indent the expressions so we've got:

```haskell
def main:
    def foo:
        none1 = None
        none1
    test = "Test me!"
```
And here is another error - after entering `foo` node we see 2 non connected nodes. No connection means there is no logical connection of the entities and here is the same variable and they need to be connected.

### Expected Result

Every step above should behave as described. In particular, creating node `def foo` should create a new empty function. Entering it and creating nodes should create indented code. Manually indenting lines should work as well.

### Luna Version

1.4

TheOnlyArtz:
You might want to merge this issue with [#1070](https://github.com/luna/luna-studio/issues/1070)

piotrMocz:
It works very nicely!

The created function (equivalent to step 2 from above) is well-formatted both in text rep and when double-clickind the node's expression. I wasn't able to get it to crash after playing with the body of the function.

Well done!

PR #1191, opened 2018-10-09
Rework adding def nodes #1093

mikusp:
### Pull Request Description

This pull request fixes adding new functions (`def fun`) inside other functions. It also changes behaviour in respect to adding incorrect functions, so that:
- `def name` will be changed to `def name:\n    None`
- `def name:` will be changed to `def name:\n    None`
- `def 999` will be changed to `def 999:\n    None` (invalid name is not automatically changed)
- `def` will be changed to `def func:\n    None` (nameless def will be given name `func`)

If incorrect function is specified in code, it will be made valid on adding the first node inside.

### Checklist


libs/luna-empire/src/Empire/ASTOps/Builder.hs
```diff
@@ -339,28 +339,48 @@ ensureHasSelf e beg = source e >>= flip matchExpr `id` \case
...
-    self <- ASTRead.cutThroughDocAndMarked =<< use (Graph.breadcrumbHierarchy . BH.self)
+ensureFunctionIsValid :: Delta -> GraphOp ()
+ensureFunctionIsValid (fromIntegral . (+4) -> indentation) = do
+    self        <- ASTRead.cutThroughDocAndMarked
+        =<< use (Graph.breadcrumbHierarchy . BH.self)
+    let whitespace = Text.replicate indentation " "
     match self $ \case
         ASGFunction n as b -> do
+            name' <- source n >>= \a -> matchExpr a $ \case
+                Invalid IR.MissingFunctionName -> return $ Just " func"
```
ljkania: Use `pure` instead. In other places too.

libs/luna-empire/src/Empire/ASTOps/Builder.hs
```diff
@@ -339,28 +339,48 @@ ensureHasSelf e beg = source e >>= flip matchExpr `id` \case
         Code.gossipLengthsChangedBy 2 =<< target e
     _ -> throwM . SelfPortNotExistantException =<< source e
 
-ensureFunctionIsValid :: GraphOp ()
-ensureFunctionIsValid = do
-    self <- ASTRead.cutThroughDocAndMarked =<< use (Graph.breadcrumbHierarchy . BH.self)
+ensureFunctionIsValid :: Delta -> GraphOp ()
+ensureFunctionIsValid (fromIntegral . (+4) -> indentation) = do
```
ljkania: Raw variables are wrong. We should have constant for indent length.

libs/luna-empire/src/Empire/ASTOps/Builder.hs
```diff
@@ -339,28 +339,48 @@ ensureHasSelf e beg = source e >>= flip matchExpr `id` \case
...
+    self        <- ASTRead.cutThroughDocAndMarked
+        =<< use (Graph.breadcrumbHierarchy . BH.self)
+    let whitespace = Text.replicate indentation " "
     match self $ \case
         ASGFunction n as b -> do
+            name' <- source n >>= \a -> matchExpr a $ \case
+                Invalid IR.MissingFunctionName -> return $ Just " func"
+                Invalid IR.InvalidFunctionName -> return Nothing
+                Invalid a                      ->
+                    error ("ensureFunctionIsValid: " <> show a)
```
ljkania: What is the reason of camel-case if this is a message?

libs/luna-empire/src/Empire/ASTOps/Builder.hs
```diff
@@ -339,28 +339,48 @@ ensureHasSelf e beg = source e >>= flip matchExpr `id` \case
...
+                _                              -> return Nothing
+            forM_ name' $ \(name :: String) -> do
+                var <- IR.var $ convert name
+                putLayer @SpanLength var $ fromIntegral $ length name
+                Just namePos <- Code.getOffsetRelativeToFile =<< source n
+                Code.insertAt namePos $ convert name
+                IR.replace var =<< source n
+                Code.gossipLengthsChangedBy (fromIntegral $ length name)
```
ljkania: This is second use of `fromIntegral $ length name`. Consider `let`

libs/luna-empire/src/Empire/ASTOps/Builder.hs
```diff
@@ -339,28 +339,48 @@ ensureHasSelf e beg = source e >>= flip matchExpr `id` \case
...
+                    =<< target n
             section' <- source b >>= \a -> matchExpr a $ \case
-                Invalid IR.MissingSection  -> return $ Just ":\n    None"
-                Invalid IR.EmptyExpression -> return $ Just "\n    None"
+                Invalid IR.MissingSection  -> return $
+                    Just $ ":\n" <> whitespace <> "None"
+                Invalid IR.EmptyExpression -> return $
+                    Just $ "\n"  <> whitespace <> "None"
```
ljkania: suffix is the same in both cases. Consider `let`

libs/luna-empire/src/Empire/ASTOps/Modify.hs
```diff
@@ -280,14 +280,16 @@ renameLambdaArg p@(Port.Projection port : []) newName lam = match lam $ \case
...
-        Code.replaceAllUses arg $ convert newName
+        Code.replaceAllUses arg oldLen $ convert newName
     ASGFunction _ as' _ -> do
         as <- ptrListToList as'
         for_ (as ^? ix port) $ \alink -> do
             arg <- source alink
+            oldLen <- getLayer @SpanLength arg
             renameVar arg newName
-            Code.replaceAllUses arg $ convert newName
+            Code.replaceAllUses arg oldLen $ convert newName
```
ljkania: These three lines are the same as in previous case. Refactor?

libs/luna-empire/src/Empire/Commands/Graph.hs
```diff
@@ -148,11 +148,21 @@ addNodeCondTC :: Bool -> GraphLocation -> NodeId -> Text -> NodeMeta -> Empire E
 addNodeCondTC tc loc@(GraphLocation f _) uuid expr meta
     | GraphLocation _ (Breadcrumb []) <- loc = do
         node <- addFunNode loc AppendNone uuid expr meta
+        when_ (node ^. Node.canEnter) $ do
+            withGraph (loc |> Breadcrumb.Definition (node ^. Node.nodeId)) $
+                let noIndent = 0
```
ljkania: <3

Issue #1098, opened 2018-09-21
Backspace breaks nodes


wdanilo:
### General Summary

Backspace in text editor works really strange sometimes. It's critical cause such operation is **very** common.

### Steps to Reproduce

1. Open new scene
2. Remove body of main and write instead (don't make any mistake, just write it):

```haskell
import Std.Base

def main:
    foo = 1 
    foo . bar
```

3. Set cursor after `bar` and press backspace unless you've got in line only `foo` left.
4. Write `. bar` again, so you get the same code as earlier. The result:

![9](https://user-images.githubusercontent.com/1623053/45883374-33293600-bdb1-11e8-8a16-329b84f8c4fa.png)

(there is **nothing** more on the stage)

### Expected Result

- Pressing backspace should not break anything.

### Luna Version
1.4

piotrMocz:
This works -- pressing backspace and writing the same thing again produces the same result. Although, I must say that in general the way the graph updates after writing in the text editor is not perfect -- after writing the node's name I need to at least enter a whitespace for the graph editor to catch up (and color the node/display its value). That is, however, known and tracked elsewhere, I believe.

PR #1201, opened 2018-10-11
Uses of vars visible as nodes

mikusp:
### Pull Request Description

This pull request fixes visibility of nodes that reference variables defined beforehand. Consider code:
```
def main:
    foo = 1
    foo
    None
```
There was only one node visible, `foo = 1`, even though second use of `foo` should be visible too. This caused some strange issues like in #1098. Such problems should now be fixed, including variables bound by pattern matches.

### Checklist

### DIFF
--- a/libs/luna-empire/src/Empire/ASTOps/BreadcrumbHierarchy.hs
+++ b/libs/luna-empire/src/Empire/ASTOps/BreadcrumbHierarchy.hs
@@ -44,13 +44,15 @@ getMarker marker = do
     matchExpr marker $ \case
         Marker index -> return index
 
-childrenFromSeq :: Delta -> EdgeRef -> GraphOp (Map NodeId BH.BChild)
-childrenFromSeq tgtBeg edge = do
+childrenFromSeq :: NodeRef -> Delta -> EdgeRef -> GraphOp (Map NodeId BH.BChild)
+childrenFromSeq currentFun tgtBeg edge = do
     ref <- source edge
     off <- Code.getOffsetRelativeToTarget edge
     let beg = tgtBeg + off
     matchExpr ref $ \case
-        Seq    l r    -> Map.union <$> (childrenFromSeq beg (coerce l)) <*> (lambdaChildren beg (coerce r))
+        Seq    l r    -> Map.union
+            <$> childrenFromSeq currentFun beg (coerce l)
+            <*> lambdaChildren  currentFun beg (coerce r)
         Marked m expr -> do
             expr'      <- source expr
             index      <- getMarker =<< source m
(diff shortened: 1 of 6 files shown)


libs/luna-empire/src/Empire/ASTOps/BreadcrumbHierarchy.hs
```diff
@@ -60,43 +62,57 @@ childrenFromSeq tgtBeg edge = do
...
+                marker <- getLayer @Marker ref
+                if isJust marker
+                then pure Map.empty
+                else childrenFromSeq currentFun tgtBeg edge
+            else do
+                tgt    <- target edge
+                seq    <- ASTRead.isSeq tgt
+                if seq
+                then childrenFromSeq currentFun tgtBeg edge
+                else pure Map.empty
```
ljkania: Consider removing duplication like this:
```
proceed <- if output == ref
    then isJust <$> getLayer @Marker ref
    else ASTRead.isSeq =<< target edge
if proceed then childrenFromSeq currentFun tgtBeg edge ele pure mampty
```

libs/luna-empire/src/Empire/ASTOps/Read.hs
```diff
@@ -414,6 +414,16 @@ isRecord expr = match expr $ \case
     ClsASG{} -> return True
     _     -> return False
 
+isNone :: NodeRef -> GraphOp Bool
+isNone = flip matchExpr $ \case
+    Cons n _ -> return $ n == "None"
+    _        -> return False
```
ljkania: Use `pure` instead

ljkania: Can we encode "None" in some variable?

libs/luna-empire/src/Empire/Commands/Graph/Metadata.hs
```diff
@@ -94,13 +94,15 @@ extractMarkedMetasAndIds
...
-        nid    <- ASTRead.getNodeId  expr
-        nid2   <- ASTRead.getNodeId  root
-        pure $ (marker, (meta, nid <|> nid2)) : rest
+        meta    <- AST.readMeta root
+        marker  <- getMarker =<< source m
+        expr    <- source e
+        rest    <- extractMarkedMetasAndIds expr
+        var     <- ASTRead.isVar expr
+        nidExpr <- ASTRead.getNodeId  expr
+        nidRoot <- ASTRead.getNodeId  root
```
ljkania: not needed extra space

libs/luna-empire/test/Test/Graph/NodeVisibilitySpec.hs
```diff
@@ -0,0 +1,94 @@
...
+                nodes      <- Graph.getNodes gl
+                parseError <- Graph.withUnit (top gl) $
+                    use $ Graph.userState . Graph.clsParseError
+                return (nodes, parseError)
+            in testCase code expectedCode $ \loc@(GraphLocation file _) -> do
+                Graph.substituteCodeFromPoints file
+                    [TextDiff (Just (Point 7 4, Point 7 4)) ".bar" Nothing]
+                (nodes, parseError) <- prepare loc
+                nodes `shouldSatisfy` (== 2) . length
+                parseError `shouldSatisfy` isNothing
```
ljkania: Consider vertical alignment for test conditions.

PR #1216, opened 2018-10-18
Erroneous markers removal

### LINKED ISSUE #1087: Breaking line breaks the project
### General Summary

Breaking lines was never supported but now we are unable to revert broken state by using undo.

### Steps to Reproduce

1. Make some line in luna-studio and break it into two lines (I did it before `.`)
2. Graph is displaying some errors
3. Press ctrl-z

### Expected Result

- Change is reverted and graph is in state before 1.

### Actual Result

- Graph is still broken but in different way.

### Luna Version

- 1.41016

mikusp:
### Pull Request Description

After each substitute, we run `sanitizeMarkers` function on code, that attempts to remove markers that are incorrect, according to lexer output, or correct but their placement does not make sense in the context of code. Fixes #1087 and #448. 

If possible, please test before accepting. I've typed successfully TramsPlayground from scratch but I'd like to see more tests for this functionality.

### Checklist

### DIFF
--- a/libs/luna-empire/src/Empire/Commands/Graph/Code.hs
+++ b/libs/luna-empire/src/Empire/Commands/Graph/Code.hs
@@ -87,7 +87,7 @@ cumulativeOffsetStream tokens = tokens' where
 -- | Takes a pair of consecutive tokens and determines if a marker
 --   is placed in a proper position. If the second token is not a marker
 --   or if it seems to be correctly placed, `Nothing` is returned.
---   If marker is determined to be incorrectly placed, Just offset
+--   If marker is determined to be incorrectly placed, `Just offset`
 --   is returned, where offset is taken from the second token.
 --   This function assumes that Tokens have cumulative offset inside
 --   them, so offset returned is offset from the beginning of the code.
--- a/libs/luna-empire/src/Empire/Commands/Graph/Code.hs
+++ b/libs/luna-empire/src/Empire/Commands/Graph/Code.hs
@@ -61,40 +61,74 @@ substituteCodeFromPoints path (breakDiffs -> diffs) = do
         pure $ map (toRealDelta . toDelta) diffs
     substituteCode path changes
 
+-- | removeMarker removes marker at given position, assuming that marker
+--   consists of '«', non-zero number of digits, and '»'.
+removeMarker :: Text -> Int -> Text
+removeMarker code markerPos =
+    let (before, after) = Text.splitAt markerPos code
+        dropMarker t =
+            let Just (markerStart, rest) = Text.uncons t
+            in if markerStart == '«'
+                then
+                    let numberDropped = Text.dropWhile isDigit rest
+                        Just (markerEnd, rest') = Text.uncons numberDropped
+                    in if markerEnd == '»'
+                        then rest'
+                        else after
+                else error $ "marker start is wrong: " <> [markerStart]
+    in Text.concat [before, dropMarker after]
+
+-- | This function computes offsets of each token from the beginning
+--   of the file and stores them in _offset field of Token
+cumulativeOffsetStream :: [Lexer.Token a] -> [Lexer.Token a]
+cumulativeOffsetStream []     = []
+cumulativeOffsetStream tokens = tokens' where
+    tokens'    = scanl1 f tokens
+    f (Lexer.Token prevSpan accOffset _) (Lexer.Token span offset lexeme)
+        = Lexer.Token (span+offset) (accOffset + prevSpan) lexeme
+
+-- | Takes a pair of consecutive tokens and determines if a marker
+--   is placed in a proper position. If the second token is not a marker
+--   or if it seems to be correctly placed, Nothing is returned.
+--   If marker is determined to be incorrectly placed, Just offset
+--   is returned, where offset is taken from the second token.
+--   This function assumes that Tokens have cumulative offset inside
+--   them, so offset returned is offset from the beginning of the code.
+--
+--   Marker is thought to be wrongly placed if it's directly preceded
+--   by anything else than:
+--       - beginning of code
+--       - end of line
+--       - block start (:)
+isWrongMarker
+    :: (Lexer.Token Lexer.Symbol, Lexer.Token Lexer.Symbol)
+    -> Maybe Delta
+isWrongMarker tokens
+    | (Lexer.Token _ _ preceding,
+       Lexer.Token _ offset (Lexer.Marker _)) <- tokens =
+        case preceding of
+            Lexer.STX        -> Nothing
+            Lexer.EOL        -> Nothing
+            Lexer.BlockStart -> Nothing
+            _                -> Just offset
+    | otherwise = Nothing
+
 sanitizeMarkers :: Text -> Text
 sanitizeMarkers text = if null erroneousMarkers
     then text
     else removeErroneousMarkers (coerce erroneousMarkers) text where
-        lexerStream  = Lexer.evalDefLexer (convert text)
-        f (Lexer.Token accS accO _) (Lexer.Token s o a) = Lexer.Token (s+o) (accO + accS) a
-        cumulativeStream = scanl f (Lexer.Token 0 0 Lexer.STX) lexerStream
-        markersIndices = findIndices (== '«') $ toList text
-        tokensForMarkers = map (\a -> (a, findToken cumulativeStream a)) $ coerce markersIndices
-        erroneousMarkers = [ a | (a, Nothing) <- tokensForMarkers] <> wrongMarkers
+        removeErroneousMarkers markers code
+            = foldl' removeMarker code (reverse $ sort markers)
+        lexerStream      = Lexer.evalDefLexer (convert text)
+        cumulativeStream = cumulativeOffsetStream lexerStream
+        markersIndices   = findIndices (== '«') $ toList text
+        tokensForMarkers = map (\a -> (a, findToken cumulativeStream a))
+            $ coerce markersIndices
+        findToken stream index = find (\t -> t ^. Lexer.offset == index) stream
+        erroneousMarkers =
+            [ a | (a, Nothing) <- tokensForMarkers] <> wrongMarkers
         precedingTokens  = zip cumulativeStream (tail cumulativeStream)
-        filterWrongMarkers a
-            | (preceding, Lexer.Token _ offset (Lexer.Marker _)) <- a = case preceding of
-                Lexer.Token _ _ Lexer.EOL        -> Nothing
-                Lexer.Token _ _ Lexer.BlockStart -> Nothing
-                Lexer.Token _ _ Lexer.STX        -> Nothing
-                _ -> Just offset
-            | otherwise = Nothing
-        wrongMarkers = catMaybes $ map filterWrongMarkers precedingTokens
-        findToken stream index = find (\(Lexer.Token _ offset _) -> offset == index) stream
-        removeErroneousMarkers markers code = foldl' removeMarker code (reverse $ sort markers)
-        removeMarker code markerPos =
-            let (before, after) = Text.splitAt markerPos code
-                dropMarker t =
-                    let Just (markerStart, rest) = Text.uncons t
-                    in if markerStart == '«'
-                        then
-                            let numberDropped = Text.dropWhile isDigit rest
-                                Just (markerEnd, rest') = Text.uncons numberDropped
-                            in if markerEnd == '»'
-                                then rest'
-                                else error $ "marker end is wrong: " <> [markerEnd]
-                        else error $ "marker start is wrong: " <> [markerStart]
-            in Text.concat [before, dropMarker after]
+        wrongMarkers     = catMaybes $ map isWrongMarker precedingTokens
 
 substituteCode :: FilePath -> [(Delta, Delta, Text)] -> Empire ()
 substituteCode path changes = do
--- a/libs/luna-empire/src/Empire/Commands/Graph/Code.hs
+++ b/libs/luna-empire/src/Empire/Commands/Graph/Code.hs
@@ -2,21 +2,24 @@ module Empire.Commands.Graph.Code where
 
 import Empire.Prelude hiding (range)
 
-import qualified Data.Map                      as Map
-import qualified Data.Text                     as Text
-import qualified Empire.ASTOps.Parse           as ASTParse
-import qualified Empire.ASTOps.Read            as ASTRead
-import qualified Empire.Commands.Code          as Code
-import qualified Empire.Commands.Publisher     as Publisher
-import qualified Empire.Data.Graph             as Graph
-import qualified Empire.Data.Library           as Library
-import qualified Empire.Data.FileMetadata      as FileMetadata
-import qualified Empire.Empire                 as Empire
-import qualified LunaStudio.Data.GraphLocation as GraphLocation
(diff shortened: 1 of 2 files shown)


libs/luna-empire/src/Empire/Commands/Graph/Code.hs
```diff
@@ -57,10 +58,79 @@ substituteCodeFromPoints path (breakDiffs -> diffs) = do
...
+        dropMarker t =
+            let Just (markerStart, rest) = Text.uncons t
+            in if markerStart == '«'
+                then
+                    let numberDropped = Text.dropWhile isDigit rest
+                        Just (markerEnd, rest') = Text.uncons numberDropped
+                    in if markerEnd == '»'
+                        then rest'
+                        else after
+                else error $ "marker start is wrong: " <> [markerStart]
```
iamrecursion: Should we be using `error` rather than something more structured?

libs/luna-empire/src/Empire/Commands/Graph/Code.hs
```diff
@@ -57,10 +58,79 @@ substituteCodeFromPoints path (breakDiffs -> diffs) = do
         pure $ map (toRealDelta . toDelta) diffs
     substituteCode path changes
 
+-- | removeMarker removes marker at given position, assuming that marker
```
iamrecursion: Would it potentially be better to employ the lexer itself to find the marker span? That way if we change what markers look like this code would still work?

libs/luna-empire/src/Empire/Commands/Graph/Code.hs
```diff
@@ -57,10 +58,79 @@ substituteCodeFromPoints path (breakDiffs -> diffs) = do
...
+--   consists of '«', non-zero number of digits, and '»'.
+removeMarker :: Text -> Int -> Text
+removeMarker code markerPos =
+    let (before, after) = Text.splitAt markerPos code
+        dropMarker t =
+            let Just (markerStart, rest) = Text.uncons t
+            in if markerStart == '«'
```
iamrecursion: Same here, can use the token definition from [`Luna.Syntax.Text.Lexer.Grammar`](https://github.com/luna/luna/blob/master/syntax/text/lexer/src/Luna/Syntax/Text/Lexer/Grammar.hs#L331), given the lexer is imported anyway?

libs/luna-empire/src/Empire/Commands/Graph/Code.hs
```diff
@@ -57,10 +58,79 @@ substituteCodeFromPoints path (breakDiffs -> diffs) = do
...
+cumulativeOffsetStream :: [Lexer.Token a] -> [Lexer.Token a]
+cumulativeOffsetStream []     = []
+cumulativeOffsetStream tokens = tokens' where
+    tokens'    = scanl1 f tokens
+    f (Lexer.Token prevSpan accOffset _) (Lexer.Token span offset lexeme)
+        = Lexer.Token (span+offset) (accOffset + prevSpan) lexeme
+
+-- | Takes a pair of consecutive tokens and determines if a marker
+--   is placed in a proper position. If the second token is not a marker
+--   or if it seems to be correctly placed, Nothing is returned.
```
iamrecursion:
```suggestion
--   or if it seems to be correctly placed, `Nothing` is returned.
```

libs/luna-empire/src/Empire/Commands/Graph/Code.hs
```diff
@@ -57,10 +58,79 @@ substituteCodeFromPoints path (breakDiffs -> diffs) = do
...
+--   If marker is determined to be incorrectly placed, Just offset
```
iamrecursion:
```suggestion
--   If marker is determined to be incorrectly placed, `Just offset`
```

libs/luna-empire/test/Test/Code/MarkerSanitizerSpec.hs
```diff
@@ -0,0 +1,134 @@
...
+            |]
+        expectedCode = normalizeLunaCode [r|
+            import Std.Base
+
+            «0»def main:
+                «1»node = foo . bar
+                None
+            |]
+        in sanitizeMarkers initialCode `shouldBe` expectedCode
+    it "removes three consecutive markers" $ let
```
iamrecursion: Does it do this if the markers are separated by spaces?

libs/luna-empire/test/Test/Code/MarkerSanitizerSpec.hs
```diff
@@ -0,0 +1,134 @@
...
+                «1»foo = 1
+                None
+            |]
+        in sanitizeMarkers initialCode `shouldBe` expectedCode
+    it "removes marker from accessor with spaces" $ let
+        initialCode = normalizeLunaCode [r|
+            import Std.Base
+
+            «0»def main:
+                «1»node = foo «2». bar
```
iamrecursion: What about if the marker is the other side of the `.` operator?

libs/luna-empire/src/Empire/Commands/Graph/Code.hs
mikusp: Sometimes lexer does not report markers. `No<1>ne` will be lexed as Invalid InvalidSuffix and there's no Marker token to look for.

libs/luna-empire/test/Test/Code/MarkerSanitizerSpec.hs
mikusp: It should, I'll add test for that.

libs/luna-empire/src/Empire/Commands/Graph/Code.hs
iamrecursion: Makes sense!

libs/luna-empire/src/Empire/Commands/Graph/Code.hs
mikusp: I think it's more like an assertion, if this case is hit, something deeply wrong happened and there's no point in trying to recover.

libs/luna-empire/src/Empire/Commands/Graph/Code.hs
```diff
@@ -1,22 +1,24 @@
...
+import qualified Empire.ASTOps.Parse                as ASTParse
+import qualified Empire.ASTOps.Read                 as ASTRead
+import qualified Empire.Commands.Code               as Code
+import qualified Empire.Commands.Publisher          as Publisher
+import qualified Empire.Data.Graph                  as Graph
+import qualified Empire.Data.Library                as Library
+import qualified Empire.Data.FileMetadata           as FileMetadata
+import qualified Empire.Empire                      as Empire
+import qualified LunaStudio.Data.GraphLocation      as GraphLocation
+import qualified Luna.Syntax.Text.Lexer             as Lexer
```
ljkania: No need for extra spaces. Please revert

libs/luna-empire/src/Empire/Commands/Graph/Code.hs
```diff
@@ -57,10 +58,79 @@ substituteCodeFromPoints path (breakDiffs -> diffs) = do
         pure $ map (toRealDelta . toDelta) diffs
     substituteCode path changes
 
+-- | removeMarker removes marker at given position, assuming that marker
+--   consists of '«', non-zero number of digits, and '»'.
+removeMarker :: Text -> Int -> Text
+removeMarker code markerPos =
+    let (before, after) = Text.splitAt markerPos code
+        dropMarker t =
+            let Just (markerStart, rest) = Text.uncons t
```
ljkania: Is this safe?

libs/luna-empire/src/Empire/Commands/Graph/Code.hs
```diff
@@ -57,10 +58,79 @@ substituteCodeFromPoints path (breakDiffs -> diffs) = do
...
+                        else after
+                else error $ "marker start is wrong: " <> [markerStart]
+    in Text.concat [before, dropMarker after]
+
+-- | This function computes offsets of each token from the beginning
+--   of the file and stores them in _offset field of Token
+cumulativeOffsetStream :: [Lexer.Token a] -> [Lexer.Token a]
+cumulativeOffsetStream []     = []
+cumulativeOffsetStream tokens = tokens' where
+    tokens'    = scanl1 f tokens
```
ljkania: Won't it work for empty list as well?

libs/luna-empire/src/Empire/Commands/Graph/Code.hs
```diff
@@ -57,10 +58,79 @@ substituteCodeFromPoints path (breakDiffs -> diffs) = do
...
+                        then rest'
+                        else after
+                else error $ "marker start is wrong: " <> [markerStart]
+    in Text.concat [before, dropMarker after]
+
+-- | This function computes offsets of each token from the beginning
+--   of the file and stores them in _offset field of Token
+cumulativeOffsetStream :: [Lexer.Token a] -> [Lexer.Token a]
+cumulativeOffsetStream []     = []
+cumulativeOffsetStream tokens = tokens' where
```
ljkania: no reason for defining `token'`. Inline instead

libs/luna-empire/src/Empire/Commands/Graph/Code.hs
```diff
@@ -57,10 +58,79 @@ substituteCodeFromPoints path (breakDiffs -> diffs) = do
...
+--   by anything else than:
+--       - beginning of code
+--       - end of line
+--       - block start (:)
+isWrongMarker
+    :: (Lexer.Token Lexer.Symbol, Lexer.Token Lexer.Symbol)
+    -> Maybe Delta
+isWrongMarker tokens
+    | (Lexer.Token _ _ preceding,
+       Lexer.Token _ offset (Lexer.Marker _)) <- tokens =
```
ljkania: can't we use lenses instead?

libs/luna-empire/src/Empire/Commands/Graph/Code.hs
```diff
@@ -57,10 +58,79 @@ substituteCodeFromPoints path (breakDiffs -> diffs) = do
...
+            Lexer.STX        -> Nothing
+            Lexer.EOL        -> Nothing
+            Lexer.BlockStart -> Nothing
+            _                -> Just offset
+    | otherwise = Nothing
+
+sanitizeMarkers :: Text -> Text
+sanitizeMarkers text = if null erroneousMarkers
+    then text
+    else removeErroneousMarkers (coerce erroneousMarkers) text where
```
ljkania: `let ... in ...` pattern would be more readable

libs/luna-empire/src/Empire/Commands/Graph/Code.hs
```diff
@@ -57,10 +58,79 @@ substituteCodeFromPoints path (breakDiffs -> diffs) = do
...
+        removeErroneousMarkers markers code
+            = foldl' removeMarker code (reverse $ sort markers)
+        lexerStream      = Lexer.evalDefLexer (convert text)
+        cumulativeStream = cumulativeOffsetStream lexerStream
+        markersIndices   = findIndices (== '«') $ toList text
+        tokensForMarkers = map (\a -> (a, findToken cumulativeStream a))
+            $ coerce markersIndices
```
ljkania:
```
tokensForMarkers = (id &&& findToken cumulativeStream) <$> coerce markersIndices
```

libs/luna-empire/src/Empire/Commands/Graph/Code.hs
```diff
@@ -57,10 +58,79 @@ substituteCodeFromPoints path (breakDiffs -> diffs) = do
...
+        findToken stream index = find (\t -> t ^. Lexer.offset == index) stream
+        erroneousMarkers =
+            [ a | (a, Nothing) <- tokensForMarkers] <> wrongMarkers
```
ljkania: <3

libs/luna-empire/src/Empire/Commands/Graph/Code.hs
```diff
@@ -57,10 +58,79 @@ substituteCodeFromPoints path (breakDiffs -> diffs) = do
...
+        precedingTokens  = zip cumulativeStream (tail cumulativeStream)
```
ljkania: `$`

libs/luna-empire/src/Empire/Commands/Graph/Code.hs
```diff
@@ -57,10 +58,79 @@ substituteCodeFromPoints path (breakDiffs -> diffs) = do
...
+        wrongMarkers     = catMaybes $ map isWrongMarker precedingTokens
```
ljkania: wrongMarkers     = catMaybes $ isWrongMarker <$> precedingTokens

libs/luna-empire/test/Test/Code/MarkerSanitizerSpec.hs
```diff
@@ -0,0 +1,134 @@
+module Test.Code.MarkerSanitizerSpec (spec) where
+
+import Empire.Prelude
+
+import Empire.Commands.Graph.Code     (sanitizeMarkers)
+import Test.Hspec                     (Spec, describe, it, parallel, shouldBe)
+import Test.Hspec.Empire              (normalizeLunaCode)
+import Text.RawString.QQ              (r)
```
ljkania: Extra spaces

libs/luna-empire/src/Empire/Commands/Graph/Code.hs
pmlodawski: Even if it is safe now:
- it generates warning, so it have to be fixed
- safeness can be changed in the future when somebody will change something in this code and forgot about this unsafe pattern match
- there is no proof of safeness in the scope, or even a comment

libs/luna-empire/src/Empire/Commands/Graph/Code.hs
pmlodawski: magic constant

libs/luna-empire/src/Empire/Commands/Graph/Code.hs
```diff
@@ -57,10 +58,79 @@ substituteCodeFromPoints path (breakDiffs -> diffs) = do
...
+removeMarker :: Text -> Int -> Text
+removeMarker code markerPos =
+    let (before, after) = Text.splitAt markerPos code
+        dropMarker t =
+            let Just (markerStart, rest) = Text.uncons t
+            in if markerStart == '«'
+                then
+                    let numberDropped = Text.dropWhile isDigit rest
+                        Just (markerEnd, rest') = Text.uncons numberDropped
+                    in if markerEnd == '»'
```
pmlodawski: magic constant

libs/luna-empire/src/Empire/Commands/Graph/Code.hs
```diff
@@ -57,10 +58,79 @@ substituteCodeFromPoints path (breakDiffs -> diffs) = do
...
+            _                -> Just offset
+    | otherwise = Nothing
+
+sanitizeMarkers :: Text -> Text
+sanitizeMarkers text = let
+    removeErroneousMarkers markers code
+        = foldl' removeMarker code (reverse $ sort markers)
+    lexerStream      = Lexer.evalDefLexer (convert text)
+    cumulativeStream = cumulativeOffsetStream lexerStream
+    markersIndices   = findIndices (== '«') $ toList text
```
pmlodawski: magic constant

Issue #1025, opened 2018-08-07
Saved projects don't update the name in .luna-project


piotrMocz:
Steps to reproduce:
1. Open a new, empty project.
2. Save it, choosing an arbitrary name other than `UnsavedLunaProject`.
3. Go to the location where you saved the project and inspect the contents of the `.luna-package` directory. You will notice two things:
    3.1. The name of the `*.lunaproject` file is still `UnsavedLunaProject`.
    3.2. The name of the project inside `config.yaml` is `UnsavedLunaProject` as well.

iamrecursion:
This should be implemented by luna/luna#258.

iamrecursion:
Functionality now implemented in https://github.com/luna/luna/pull/260, waiting for review.

mikusp:
@iamrecursion I used this `rename` function instead of my previous implementation, and see that it cannot be done using `Dir.renameDirectory` and `Dir.renameFile`. I get an exception `rename: unsupported operation (Invalid cross-device link)`. You should copy the files and then delete original directory.

iamrecursion:
Will fix when back!

iamrecursion:
Fixed [here](https://github.com/luna/luna/commit/588c2519b3bdcc75e43586256a707a3e3ddf8d19). Needs to be integrated with Studio now.

PR #1219, opened 2018-10-19
Use rename from luna #1025

mikusp:
### Pull Request Description

Use logic of renaming projects from `luna`

### Checklist


libs/batch/plugins/luna-empire/src/Empire/Server/Atom.hs
```diff
@@ -88,22 +89,10 @@ partitionM f (x:xs) = do
...
-                Dir.createDirectoryIfMissing True $ newPath </> relPath
-            forM_ filesToCopy $ \f -> do
-                let relPath = makeRelative oldPath f
-                Dir.copyFile f (newPath </> relPath)
-        Dir.removeDirectoryRecursive oldPath `catchAny` (\e -> logger Logger.error $ "couldn't remove directory" <> oldPath)
-        return r
+    result <- liftIO $ try $ do
+        src    <- Path.parseAbsDir oldPath
+        target <- Path.parseAbsDir newPath
+        Package.rename src target
```
iamrecursion: I don’t like the catch all reflexively, but I don’t think there’s any point in more granular catches either.

Issue #912, opened 2018-05-08
Sort out coordinates directions for keyboard shortcuts


kustosz:
Up arrow should move up, not down.

sylwiabr:
on a graph like this:
![obraz](https://user-images.githubusercontent.com/12892578/47736967-1ca7b180-dc70-11e8-9ed6-476b5e5c4850.png)
select node `rolling2` press right arrow -> I would suspect going to node `sum1` and I am getting to node `at1`
If this would be correct than from node `at1` pressing left arrow I should go to node `rolling2` but I am goint to node `rolling1` the code for the project is: 
```
import Dataframes.Internal.CWrappers
import Dataframes.Learn.SKLearnWrapper
import Dataframes.Column
import Dataframes.Table
import Dataframes.Main
import Dataframes.IO
import Std.Base
import Std.Foreign.C.Value
import Std.Foreign

import Dataframes.Plot.Matplotlib

def abs name x:
    (x.at name < 0) . switch (-(x.at name)) (x.at name)

def filterOutlayers filledData colName:
    at1 = filledData . column colName
    meanValue1 = at1 . mean . toList . head . get
    stdValue1 = at1 . std . toList . head . get
    at2 = filledData.at colName
    eachTo1 = at2 . eachTo "less_mean" (_.at colName - meanValue1 )
    eachTo2 = eachTo1 . eachTo "abs" (abs "less_mean")
    eachTo3 = eachTo2 . eachTo "filtered" (x: (x.at "abs" < 3.0 * stdValue1 ).switch (x.at colName) meanValue1 )
    column1 = eachTo3 . column "filtered"
    column1

def bollingerBands sum1:
    eachTo1 = sum1 . eachTo "close_std_2" (_.at "Close_std dev" * 2.0)
    eachTo2 = eachTo1 . eachTo "BB_Lower" (x: x.at "Close_mean" - x.at "close_std_2")
    eachTo3 = eachTo2 . eachTo "BB_Upper" (x: x.at "Close_mean" + x.at "close_std_2")
    eachTo3

def timeseries:
    read1 = Table.fromFile "/home/sylwia/luna/projects/Dataframes/data/Stocks/medp.us.txt"
    rolling1 = read1 . rollingInterval  "Date" 20.days [("Close", [Std])]
    rolling2 = read1 . rollingInterval "Date" 20.days [("Close", [Mean])]
    at1 = rolling1 . at "Close_std dev"
    sum1 = rolling2 . + at1
[...]

PR #1225, opened 2018-10-23
Reverse y-axis in graph navigation. #912

pmlodawski:
### Pull Request Description

- As new gui has reversed y axis, we need also to reverse graph navigation using keyboard shortcuts

### Checklist


ljkania:
Looks good but it needs to be very well tested. Are you sure these changes take care of everything? (like selection box for example)

pmlodawski:
We have no selection box yet in new gui, so yes ;)
This bug is only about navigation and this works fine to me.

PR #1252, opened 2018-11-06
Initial fixes

wdanilo:
### Pull Request Description

The quality of the code in build scripts is **terrible**. In particular:

- Global paths are changed between imports of modules and modules depend on it. (Yup, there is code running between importing python modules which glue things).
- Path management is in a module "atom_prepare" and it is imported by all other modules, like modules managing stack, which does not have anything in common with atom.
- Path management is created in a way, that every hardcoded path in code starts with "../" because the utils hardcode some internal folder path to start with so first you need to go level up. It's literally everywhere.
- Try...catch is used instead of if...then...else. In particular, there are catches which catch just everything and print "FAIL", which makes debugging impossible.
- The names, like "luna-studio" are hardcoded in many, many places across files. It needs to be refactored.
- The overall quality of code is very poor, instead of suing Pythons *args and **kwargs mechanisms, the variables are deconstructed to multiple variables, passed separately to functions and adding any new build parameter requires to change many places in the code.
- In the repo, there are some binary files, without description and they are just evaluated? during compilation on Windows. It seems like a bad design. 
- And many more.

I've rewritten some part of it to be able to understand what the scripts actually do. Anyway, there is needed a big rewrite to it, as the quality in every other module is very, very poor.


wdanilo:
Another bad thing is that build scripts are modifying files in source directory. This makes the whole process error prone and very complex. It needs to be resolved asap.

sylwiabr:
Can we do an issue with this instead of PR @wdanilo @mwu-tow

wdanilo:
@mwu-tow pls take a look here.

Issue #1254, opened 2018-11-07
Cannot save TramsPlayground example


mwu-tow:
I use luna-studio  1.5.100.94 

### Steps to Reproduce
1. Open the TramsPlayground example
1. Move focus to node editor
1. Presss Ctrl+S
1. Accept default path by Enter — I don't have anything in my filesystem under proposed path

### Expected Result
Project is saved at given path.

### Actual Result
Error appears:
```
C:\Users\mwurb\AppData\Local\Temp\luna13357\tutorials\TramsPlayground\: removeDirectoryRecursive:removeContentsRecursive:RemoveDirectory "\\\\?\\C:\\Users\\mwurb\\AppData\\Local\\Temp\\luna13357\\tutorials\\TramsPlayground\\": unsatisfied constraints (ディレクトリが空ではありません。)
CallStack (from -prof):

while processing event

empire.atom.project.move.response

======= logs: =======

=== s-luna-empire.err.log ===
ERROR 2018-11-07T13:02:53.8122235Z	:: error during processing request empire.atom.project.move.request: C:\Users\mwurb\AppData\Local\Temp\luna13357\tutorials\TramsPlayground\: removeDirectoryRecursive:removeContentsRecursive:RemoveDirectory "\\\\?\\C:\\Users\\mwurb\\AppData\\Local\\Temp\\luna13357\\tutorials\\TramsPlayground\\": unsatisfied constraints (ディレクトリが空ではありません。)
CallStack (from -prof):


=== s-luna-empire.out.log ===


=== s-luna-empire.wrapper.log ===
2018-11-07 14:02:17,700 INFO  - Starting ServiceWrapper in the CLI mode
2018-11-07 14:02:18,129 INFO  - Completed. Exit code is 0
```

The error message says that directory `C:\Users\mwurb\AppData\Local\Temp\luna13357\tutorials\TramsPlayground` is not empty. 
After I checked it later the directory exists and is empty.

Also, for some reason text editor and nodes disappeared. I cannot continue work (and lost any previous work that might have been done).
Also, the path where save was attempted, seemingly contains the project. I believe it shouldn't — if storing project fails, the directory should be restored to previous state. Now when I try to repeat this error, it fails because target path is already occupied.

### Luna Version

- Please include the output of `luna-studio --version`.
The command fails with the following:
![obraz](https://user-images.githubusercontent.com/1548407/48133165-3495bb80-e296-11e8-8c16-06120710d743.png)

[...]

PR #1274, opened 2018-11-15
Do not destroy text editor when file is externally deleted (fixes bug on windows)

pmlodawski:
### Pull Request Description

Safe save mechanism is moving original file as .backup, new file is saved and then moved to original file location. This mechanism triggers `did-delete` event on windows and editor panel is destroyed which has further consequences, causing https://github.com/luna/luna-studio/issues/1254

To fix this issue, we can
1. Remove delete detection of files by external program (not *luna-studio* itself), **which this PR implements**
1. do not safe save, or fix this mechanism to not trigger `did-delete` event, **this PR does NOT implements this**

### Important Notes

As described above, **this PR introduces new bug**: external deletion of files is not detected and gui is left then in wrong state (it is not working properly until file is manually closed)

### Checklist

### DIFF
--- a/luna-studio/atom/lib/luna-code-editor-tab.coffee
+++ b/luna-studio/atom/lib/luna-code-editor-tab.coffee
@@ -49,8 +49,6 @@ TextBuffer::subscribeToFileOverride = (codeEditor) ->
         @emitter.emit 'did-delete'
         if modified
             @updateCachedDiskContents()
-        else
-            @destroy()
 
     @fileSubscriptions.add @file.onDidRename =>
         @emitter.emit 'did-change-path', @getPath()


ljkania:
What are the consequences of this PR? Does it introduce any regression or potential bugs? Please, confirm that everything will be fine or create issue for remaining problems

mwu-tow:
While I don't really understand the consequences of this PR, I can confirm that with package 1.5.100.116 the issue #1254 seems to be finally gone.

mikusp:
It seems that #1275 is a better alternative.

mwu-tow:
Please disregard my previous comment — the package I tested was with #1275, not this PR changes.

pmlodawski:
Cool! @mikusp I'm happy you found a way to fix this in backend!

PR #1307, opened 2018-12-06
Fixes for undoing node actions from code and vice versa

mikusp:
### Pull Request Description

Replaces default atom behaviour of undoing changes in code by sending whole file as Substitute request and unifies it with out undo/redo infrastructure.

### Important Notes

@pmlodawski @ljkania I have replaced generating client UUID with zeros, because this caused accidental breakage in undo - node editor was considered to be different client than code editor. Ideally, this ID should be generated before and used in these two places. Please comment if you see a way to do that

### Checklist

### DIFF
--- a/luna-studio/node-editor/src/Main.hs
+++ b/luna-studio/node-editor/src/Main.hs
@@ -2,10 +2,10 @@
 module Main where
 
 import           Common.Prelude
+import           Common.ClientId            (clientId)
 import           Control.Concurrent.Chan    (Chan)
 import qualified Control.Concurrent.Chan    as Chan
 import           Control.Concurrent.MVar
-import Data.UUID (nil)
 import qualified JS.Mount                   as Mount
 import           JS.UUID                    (generateUUID)
 import           NodeEditor.Event.Engine    (LoopRef (LoopRef))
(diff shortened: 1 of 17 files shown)


luna-studio/node-editor/src/Main.hs
```diff
@@ -21,14 +22,14 @@ import           WebSocket                  (WebSocket)
 runApp :: Chan (IO ()) -> WebSocket -> IO ()
 runApp chan socket = do
     random         <- newStdGen
-    clientId       <- generateUUID
+    -- clientId       <- generateUUID
```
pmlodawski: Why?

luna-studio/text-editor/src/Main.hs
```diff
@@ -17,11 +18,11 @@ import           WebSocket               (WebSocket)
 runApp :: Chan (IO ()) -> WebSocket -> IO ()
 runApp chan socket = do
     random       <- newStdGen
-    clientId             <- generateUUID
+    -- clientId             <- generateUUID
```
pmlodawski: Remove it if its not needed

luna-studio/node-editor/src/Main.hs
mikusp: As I said in the description, now code editor and node editor have different clientIds, so undo treats them as different users. That means that if you change code, you cannot undo that in node editor, as instead your last action in node editor will be undone. This is confusing for users.
If you see a way to generate one ID for both code and node editors, I'm all for it.

pmlodawski: I'm going to merge text-editor and code-editor in the future, can you hardcode single uuid in some common library and use in both this components?

mikusp: Sure :+1:

PR #1315, opened 2018-12-12
Retrying the frontend build

piotrMocz:
### Pull Request Description

The frontend build likes to fail randomly. While we wait for the LTS update or removing GHCJS, here's a change that will retry the frontend build a given number of times (3 by default, but you can override it with a command line argument).

### Checklist

- [] The code has been tested where possible.


pmlodawski:
:heart:

PR #1333, opened 2019-02-21
Make bootstrap describe its dependencies

mwu-tow:
### Pull Request Description
Thanks to this bootstrap.hs can be "just" executed by stack tool, without any additional knowledge on the caller site.
See also https://github.com/luna/luna-manager/pull/232

### Important Notes
Resolver has been bumped to lts-11. Anything newer won't work because of shelly. Stack script command doesn't allow for extra-deps. Creating a custom shapshot for this feels like an overkill. Should be addressed in the next set of changes to bootstrap/build scripts.

### Checklist

### DIFF
--- a/bootstrap.hs
+++ b/bootstrap.hs
@@ -1,4 +1,3 @@
-#!/usr/bin/env stack
 {- stack
   script
   --resolver lts-11.22
--- a/bootstrap.hs
+++ b/bootstrap.hs
@@ -1,3 +1,12 @@
+#!/usr/bin/env stack
+{- stack
+  script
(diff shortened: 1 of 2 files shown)


mwu-tow:
Doesn't work on macOS, will need to remove the shebang line before merge.

iamrecursion:
The shebang line is primarily for Linux. In general you can just `stack foo.hs` instead.

mwu-tow:
That's how we call it everywhere. The problem is that mere presence of shebang causes compilation failure: https://ghc.haskell.org/trac/ghc/ticket/6132

PR #1334, opened 2019-02-21
Remove installing happy from bootstrap

mwu-tow:
### Pull Request Description
There are several problems with the way we install happy:
* stack tries to use inherited stack.yaml context leading to contradictory requiremenets and build failures
* we modify environemnt outside the sandbox (`~/.local/bin`)
* we have no guarantee that `~/.local/bin` is on the PATH and the PATH prepend we do doesn't affect anyone but us (and we exit right after that)

The good news is that happy is not needed after all. 
This PR removes steps code that was installing happy as part of bootstrap routine.

### Important Notes

- Mention important elements of the design.
- Mention any notable changes to APIs. 

### Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -25,12 +25,7 @@ For a tutorial, documentation, explanation of the ideas behind Luna please visit
 ## Hacking on Luna Studio
 Luna's build system is nice and simple, allowing you to bootstrap the compiler
 as long as you have an installation of 
-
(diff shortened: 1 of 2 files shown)


mwu-tow:
Good idea. 
Actually having README state that bootstrap requires happy to be installed by user makes the whole thing even more strange.

iamrecursion:
Yeah I've forever been confused as to why that is.

PR #1335, opened 2019-02-22
 [win] refer to sc.exe by absolute path 

mwu-tow:
### Pull Request Description
Installer on Windows calls `installAll.bat` script — that in turn calls to `sc.exe` (Service Control Manager). It is located in `C:\Windows\system32`. This location, while being part of PATH by default, may be removed by user. In such case the script will fail to find the executable.

This PR fixes that by referring to `sc.exe` by its absolute path. `%SystemRoot%` built-in variable is used to detect location of Windows installation (it doesn't necessarily has to be on `C:\`). 

### Important Notes
None.

### Checklist


mwu-tow:
The directory structure is the same, regardless of architecture. Despite the name, on 64-bit Windows System32 contains 64-bit binaries. Well, given how much world depends on this path changing it wasn't feasible.
(interestingly, when running 32-bit process, it will see 32-bit binaries there thanks to some magic system-level redirecting)

We only build 64-bit binaries and target 64-bit Windows and that is what I have tested.

iamrecursion:
Brilliant!

PR #1341, opened 2019-03-05
New searcher

kustosz:
### Pull Request Description

Complete overhaul of the searcher engine. It is currently purely JS, you can see the JS library at https://github.com/luna/fuzzly. Paired with it there's a lot of changes and refactoring of the related logic in GUI, making it more engine agnostic and better in general.

### Important Notes

There shouldn't be much changes compared to the old searcher, other than much improved performance and possible slight deviations in resulting scores.

### Checklist


luna-studio/lib/src/Common/Debug.hs
```diff
@@ -16,8 +17,9 @@ import           Data.Time.Clock         (diffUTCTime, getCurrentTime)
 class HasRequestTimes st where
     requestTimes :: Lens' st (Map UUID UTCTime)
 
-measureResponseTime :: (HasRequestTimes s, Topic.MessageTopic (Response.Response req inv res)) =>
-                       Response.Response req inv res -> Command s ()
+measureResponseTime
```
iamrecursion: I would force both this and `timeAction` below to be inlined in the no debug case. That way we ensure they have no performance overhead when debugging is off.

luna-studio/lib/src/Common/Debug.hs
```diff
@@ -26,8 +28,25 @@ measureResponseTime resp = do
         Just reqTime -> do
             currTime <- getCurrentTime
             let timeDiff = show $ diffUTCTime currTime reqTime
-            putStrLn $ "[Request time -- NodeEditor] " <> Topic.topic resp <> " took " <> timeDiff
+            putStrLn $ "[Request time -- NodeEditor] " <> Topic.topic' resp <> " took " <> timeDiff
         Nothing      -> putStrLn $ "[Request time -- NodeEditor] request uuid doesn't match any known requests."
 #else
 measureResponseTime _ = return ()
```
iamrecursion:
```suggestion
measureResponseTime _ = pure ()
{-# INLINE measureResponseTime #-}
```

luna-studio/lib/src/Common/Debug.hs
```diff
@@ -26,8 +28,25 @@ measureResponseTime resp = do
...
+#ifdef DEBUG_PERF
+timeAction actName act = withLevel $ \l -> do
+    startTime <- liftIO performanceNow
+    result    <- act
+    endTime   <- liftIO performanceNow
+    let timeDiff = endTime - startTime
+    putStrLn $ replicate l ' ' <> actName <> " took: " <> show timeDiff <> "ms"
+    pure result
+#else
+timeAction _ = id
```
iamrecursion:
```suggestion
timeAction _ = id
{-# INLINE timeAction #-}
```

luna-studio/lib/src/Searcher/Data/Database.hs
```diff
@@ -0,0 +1,40 @@
...
+    jsCreate :: JSVal -> JSVal
+
+create :: SearcherData a => [a] -> Database a
+create hints = let
+    len         = length hints
+    hintsArr    = Array.listArray (0, len - 1) hints
+    assocs      = (_2 %~ view text) <$> Array.assocs hintsArr
+    -- Yes, it's an `unsafePerformIO`. It is a pure computation
+    -- by the virtue of its usage and semantics of the JS library,
+    -- so we ruthlessly navigate around GHCJS marshaling constraints.
```
iamrecursion: I assume this is purely for performance?

luna-studio/lib/src/Searcher/Engine.hs
```diff
@@ -0,0 +1,34 @@
...
+foreign import javascript safe "$1.query($2)"
+    jsQuery :: JSVal -> JSVal -> JSVal
+
+-- | Yes, there's an `unsafeFromJust` here. This is translating the results
+--   from JS FFI, if there's a `Nothing` returned, that means API mismatch
+--   and the whole world is on fire and we want to fail soon.
+--   Also yes, there's an `unsafePerformIO`. The returned value will
+--   never be mutated, thus we can do it.
+castResults :: JSVal -> [([Int], [Int], Double)]
+castResults = unsafeFromJust . unsafePerformIO . fromJSVal
```
iamrecursion: Rather than using unsafeFromJust I would suggest unpacking with a more descriptive call to error. I agree we want to fail immediately, but I think that the more information we have the better.

luna-studio/node-editor-view/src/NodeEditor/React/View/ExpressionNode.hs
```diff
@@ -325,12 +334,17 @@ nodeContainer = React.defineView name $ \(ref, isTopLevel, performConnect, maySe
...
+    -> Maybe Searcher.Properties
 filterOutSearcherIfNotRelated _  Nothing  = Nothing
-filterOutSearcherIfNotRelated nl (Just s) = if Searcher.isSearcherRelated nl s then return s else Nothing
-
-filterOutEditedTextControlIfNotRelated :: NodeLoc -> Maybe InPortRef -> Maybe InPortRef
-filterOutEditedTextControlIfNotRelated _  Nothing        = Nothing
-filterOutEditedTextControlIfNotRelated nl (Just portRef) = if isPrefixOf (toNodeIdList nl) (toNodeIdList $ portRef ^. PortRef.nodeLoc)
-    then Just portRef
+filterOutSearcherIfNotRelated nl (Just s) = if Searcher.isRelated nl s
+    then return s
```
wdanilo: General request – when compiling every module, let's watch the warnings. Basically, the new code in this PR generates a lot of warnings, can we please fix them, at least in the obvious places in old modules and in every place in the new modules?

luna-studio/node-editor-view/src/NodeEditor/React/View/Searcher.hs
```diff
@@ -104,29 +107,34 @@ results_ ref selected results = if null results then return () else
...
-    highlighted_' :: Int -> [Range] -> ReactElementM ViewEventHandler ()
+    highlights = result ^. match . wrapped
+    name'      = convert $ result ^. SearcherData.text
+    highlighted_' :: Int -> [Int] -> ReactElementM ViewEventHandler ()
     highlighted_' omit [] = span_ [ "key" $= "l" ] $ elemString $ snd $ splitAt omit name'
-    highlighted_' omit ((start, end):rest) = do
-        let len                   = end - start
+    highlighted_' omit (h:rest) = do
+        let start                 = h
+            len                   = 1
```
wdanilo: The name of this value should be more descriptive. I know it's len of 1, but what is this exaclty. Its one-letter-highlight or what is exactly this case about?

luna-studio/node-editor-view/src/NodeEditor/React/View/Sidebar.hs
```diff
@@ -14,37 +14,36 @@ import qualified NodeEditor.Event.UI                       as UI
...
-                                                  handleMouseUp, handlers,
-                                                  jsShow2, modeClass)
-import NodeEditor.React.View.Searcher            (searcher_)
-import NodeEditor.React.View.Style               (plainPath_, plainRect_)
+import Data.Timestamp                          (Timestamp (Timestamp))
+import JS.Scene                                (inputSidebarId, outputSidebarId)
+import LunaStudio.Data.PortRef                 (AnyPortRef (OutPortRef'),
+                                                OutPortRef (OutPortRef),
+                                                toAnyPortRef)
+import LunaStudio.Data.Position                (y)
```
wdanilo: come on :D I know it was this way, but it just hurts my eyes.

luna-studio/node-editor-view/src/NodeEditor/React/View/Sidebar.hs
```diff
@@ -245,11 +251,16 @@ portLabelId = Mount.prefix "focus-portLabel"
...
-        Searcher.PortNameMode d -> let sPortRef = OutPortRef (ns ^. Searcher.nodeLoc) (d ^. Searcher.portId)
-            in if portRef == sPortRef then Just s else Nothing
-        _ -> Nothing
-    _                            -> Nothing
+filterOutSearcherIfNotRelated :: AnyPortRef -> Maybe Searcher.Properties 
+    -> Maybe Searcher.Properties
+filterOutSearcherIfNotRelated (OutPortRef' portRef) (Just s) 
+    = case s ^. Searcher.searcher . Searcher.mode of
+        SearcherMode.Node ns -> case ns ^. NodeSearcher.mode of
+            NodeSearcher.PortNameMode d -> let
```
wdanilo: what is `d` ?

luna-studio/node-editor-view/src/NodeEditor/React/View/Sidebar.hs
```diff
@@ -245,11 +251,16 @@ portLabelId = Mount.prefix "focus-portLabel"
...
-    Searcher.NodeSearcher ns -> case ns ^. Searcher.modeData of
-        Searcher.PortNameMode d -> let sPortRef = OutPortRef (ns ^. Searcher.nodeLoc) (d ^. Searcher.portId)
-            in if portRef == sPortRef then Just s else Nothing
-        _ -> Nothing
-    _                            -> Nothing
+filterOutSearcherIfNotRelated :: AnyPortRef -> Maybe Searcher.Properties 
+    -> Maybe Searcher.Properties
+filterOutSearcherIfNotRelated (OutPortRef' portRef) (Just s) 
+    = case s ^. Searcher.searcher . Searcher.mode of
+        SearcherMode.Node ns -> case ns ^. NodeSearcher.mode of
```
wdanilo: I know `ns` is short and used everywwhere, but in new code we should use more descriptive names. Its not so bad to type a few letters more and make easier time for the reader to encode what it means.

luna-studio/node-editor/src/NodeEditor/Action/Basic/UpdateSearcherHints.hs
```diff
@@ -2,169 +2,199 @@ module NodeEditor.Action.Basic.UpdateSearcherHints where
...
+updateClassName cl = do
+    modifySearcher $ do
+        Searcher.mode . Mode._Node . NodeMode.mode
+            . NodeMode._ExpressionMode . NodeMode.parent .= cl
+    updateHintsPreservingSelection
+
+scoreTextMatch :: Text -> NodeHint.Database -> [Result NodeHint.Node]
+scoreTextMatch query nsData = case Text.null query of
+    True ->
+        let mkResult r = Result r 0 $ Match []
```
wdanilo: Please, do not use hardcoded literals in expressions. Either refactor it to a well described variable or create some `mempty` - like instance to hide it in a well-described abstraction.

luna-studio/node-editor/src/NodeEditor/State/Global.hs
```diff
@@ -74,17 +73,16 @@ makeLenses ''DebugState
...
+    {- react               -} (UI.mkState ref)
+    {- backend             -} (BackendState def clientId')
+    {- actions             -} def
+    {- collaboration       -} def
+    {- debug               -} def
+    {- selectionHistory    -} def
+    {- searcherDatabase    -} def
+    {- preferedVisualizers -} mempty
+    {- visualizers         -} mempty
+    {- internalVisualizers -} mempty
```
wdanilo: omg

luna-studio/lib/src/Common/Debug.hs
kustosz: Inlining as per @iamrecursion suggestion. As for further modifications, to my knowledge (albeit limited, not an author of this) this helper is used at most once per request, which means there's a whole bunch of serializing and networking going on with each of these, meaning that a single unoptimized `pure ()` will not have measurable impact on anything.

luna-studio/lib/src/Searcher/Engine.hs
kustosz: `fromJSVal` uses `IO` because array is a reference type, meaning it can be mutated at any point. With this one, we're sure we're the owners of the returned array, so the timing of this operation does not matter.

luna-studio/lib/src/Searcher/Data/Database.hs
```diff
@@ -0,0 +1,40 @@
...
+    jsValAssocs = unsafePerformIO $ toJSVal assocs
```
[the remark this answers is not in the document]
kustosz: I am sure, from the unfortunate experience of having to read through all of GHCJS marshalling sources back when battling the serialization performance :joy:.
It's just a function that creates an empty JS Array and does a bunch of appends, all in local vars. In principle, it could be a `pToJSVal` (i.e. the pure counterpart), but to my understanding they try to have only `pToJSVal` when a `pFromJSVal` also exists, and `pFromJSVal` for arrays obviously cannot exist.

luna-studio/node-editor-view/src/NodeEditor/React/Model/Searcher/Mode.hs
```diff
@@ -1,45 +1,28 @@
...
+------------------
 
 
+-- === Definition === --
+
 data Mode
-    = CommandSearcher [Match Command]
-    | NodeSearcher    (NodesData Symbol)
+    = Command
+    | Node    Node
```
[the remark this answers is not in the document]
kustosz: `Hint` is too general, if I'm searching for commands I am also seeing hints. I like `Node` since it's very clear even though maybe not 100% correct. Maybe `Expression`? Or `Function`?

luna-studio/node-editor-view/src/NodeEditor/React/Model/Searcher/Hint/Node.hs
```diff
@@ -0,0 +1,171 @@
...
+
+----------------------
+-- === Database === --
+----------------------
+
+
+-- === Definition === --
+
+data Database = Database
+    { _database :: Database.Database Node
```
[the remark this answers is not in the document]
kustosz: Just noticed this is 6 occurences of the word `data`, 5 of `database` in 2 lines :joy:. 
Anyway, I'm changing the qual import so that it's `Searcher.Database`. As for the name types themselves, it is only fair. They both are databases and one of them is using the other. One comes from a library-ish, general part, the other is more domain specific and has more fields. No idea for a better name.

luna-studio/node-editor/src/NodeEditor/State/Global.hs
kustosz: This is a convention I see upheld throughout GUI code and to be honest it seems pretty good to me (apart from "hey, maybe let's not use types with 23 fields", but refactoring that is well beyond the scope of this PR). You only have 3 options to do this:
1. No comment – good luck figuring out.
2. Comments like this – silly but works.
3. Record syntax – good luck adding a field and forgetting to set it. With record syntax GHC does not check if all fields are set, happily using `undefined` as the value.

luna-studio/node-editor-view/src/NodeEditor/React/View/Searcher.hs
kustosz: Good point, forgot to bring range-grouping back.

libs/luna-empire/src/Empire/Commands/Graph.hs
```diff
@@ -10,110 +10,111 @@ import Empire.Commands.Graph.NodeMeta   as X
 
 import Empire.Prelude hiding (head, toList)
```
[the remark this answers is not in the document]
kustosz: To my understanding it was Marcin Mikołajczyk's effort to mitigate the pain of Prologue changing rapidly, forcing him to migrate often. If that's an effort you find worthwhile, we can purge it as a separate task.

luna-studio/lib/src/Common/Prelude.hs
```diff
@@ -41,8 +41,8 @@ import           GHCJS.Types               as X (JSVal)
 import           Prelude                   hiding (curry, error, print, putStr, putStrLn, uncurry, (++), (.))
 import           Prologue                  as X (FromList, IsList, Item, Mempty, NFData, Semigroup, ToList, convert, curry, fmap1, fmap2,
                                                  fmap3, fmap4, fmap5, fromJustM, fromList, lift2, lift3, pprint, putStr, switch, toList,
-                                                 toString, uncurry, unlessM, unwrap, whenLeft, whenM, whenM_, whenRight, withJust, wrap,
-                                                 ($>), (.), (.:), (.:.), (.::), (.::.))
+                                                 toString, uncurry, unlessM, unsafeFromJust, unwrap, whenLeft, whenM, whenM_, whenRight, withJust, wrap,
+                                                 wrapped, ($>), (<<$>>), (.), (.:), (.:.), (.::), (.::.))
```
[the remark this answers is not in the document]
kustosz: This one is trickier, for 2 reasons:
1. This prelude exports some GHCJS specific stuff.
2. New Prologue does not and probably will not compile with GHCJS. This means it's effectively branching off from the current state of prologue anyway. 
Again, we can purge it in a separate effort. I would not advise this way though, given the plan to get rid of HS bits in the GUI anyway.

luna-studio/node-editor-view/src/NodeEditor/React/Model/Searcher/Input.hs
```diff
@@ -1,3 +1,4 @@
+{-# LANGUAGE Strict #-}
```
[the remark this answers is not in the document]
kustosz: I am pretty sure global Strict would break GUI in some places, glancing at the code. We can make the effort to change it, but I would estimate it taking a couple days to find all the things that broke and fixing them.

PR #1353, opened 2019-03-20
Paren closer

kustosz:
### Pull Request Description

This PR brings automatic parentheses closing capability to searcher.

### Important Notes

1. Design is based on the behavior of analogous feature in Atom and VS Code. In particular:
    - Typing `(` before line end, whitespace or some particular characters (`:`, `)`, `]`) results in inserting `)` after the cursor.
    - In a situation like `(<cursor>)`, pressing backspace removes both opening and closing paren.
    - Typing `)` before a closing paren results in cursor jumping after the paren, without typing anything.

2. There was some obsolete event handling code that looked very promising but did nothing, other than wasting a couple hours of my time. It's deleted in this PR. Nothing changed in the behavior after this.

3. JS is not very good in handling undo-redo in inputs in the presence of programatic input changes. It just doesn't take them into account. This forced me to introduce a custom undo-redo system. It sometimes behaves counter-intuitively wrt. cursor position after undo. It is not easily avoidable in the current architecture (GUI uses an old react version, without the events necessary to handle all modes of cursors movement). This, however, should not annoy the user in most scenarios. If it does, there are additional steps we can take (i.e. reimplementing _all_ cursor movement modes, like arrow keys or Home/End, in custom code), but after consultation with @wdanilo we've decided not to do it now.

### Checklist

### DIFF
--- a/luna-studio/node-editor-view/src/NodeEditor/React/Event/Searcher.hs
+++ b/luna-studio/node-editor-view/src/NodeEditor/React/Event/Searcher.hs
@@ -21,6 +21,12 @@ data Event = InputChanged Text Int Int
            | MoveLeft
            | KeyDown KeyboardEvent
            | KeyUp   KeyboardEvent
-            deriving (FromJSON, Generic, NFData, Show, Typeable)
+           | ParenOpen
+           | ParenClose
+           | Backspace
+           | Undo
+           | Redo
+           | SelectionChanged
+           deriving (FromJSON, Generic, NFData, Show, Typeable)
 
 instance EventName Event
--- a/luna-studio/node-editor-view/src/NodeEditor/React/Event/Searcher.hs
+++ b/luna-studio/node-editor-view/src/NodeEditor/React/Event/Searcher.hs
@@ -10,23 +10,24 @@ import           Data.Aeson        (FromJSON)
 import           React.Flux        (KeyboardEvent)
 
 
-data Event = InputChanged Text Int Int
-           | Accept
-           | AcceptInput
-           | HintShortcut   Int
-           | AcceptWithHint Int
-           | TabPressed
-           | MoveDown
-           | MoveUp
-           | MoveLeft
-           | KeyDown KeyboardEvent
-           | KeyUp   KeyboardEvent
-           | ParenOpen
-           | ParenClose
-           | Backspace
-           | Undo
-           | Redo
-           | SelectionChanged
-           deriving (FromJSON, Generic, NFData, Show, Typeable)
+data Event
+    = InputChanged Text Int Int
+    | Accept
+    | AcceptInput
+    | HintShortcut   Int
+    | AcceptWithHint Int
+    | TabPressed
+    | MoveDown
+    | MoveUp
+    | MoveLeft
+    | KeyDown KeyboardEvent
+    | KeyUp   KeyboardEvent
+    | ParenOpen
+    | ParenClose
+    | Backspace
+    | Undo
+    | Redo
+    | SelectionChanged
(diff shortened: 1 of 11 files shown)


luna-studio/node-editor-view/src/NodeEditor/React/Event/Searcher.hs
```diff
@@ -21,6 +21,12 @@ data Event = InputChanged Text Int Int
            | MoveLeft
            | KeyDown KeyboardEvent
            | KeyUp   KeyboardEvent
-            deriving (FromJSON, Generic, NFData, Show, Typeable)
```
iamrecursion: We might as well refactor the formatting on this whole ADT to match the coding guidelines given nearly half of it has changed:

```hs
data Event
    = InputChanged Text Int Int
    | Accept
    | ...
    deriving (FromJSON, Generic, NFData, Show, Typeable)
```

luna-studio/node-editor-view/src/NodeEditor/React/Model/Searcher/UndoRedo.hs
```diff
@@ -0,0 +1,74 @@
...
+
+------------------------
+-- === InputState === --
+------------------------
+
+-- === Definition === --
+
+data InputState = InputState
+    { _text :: Text
+    , _selectionStart :: Int
```
iamrecursion: Might as well make these strict. There's no chance we're _not_ going to use them.

luna-studio/node-editor-view/src/NodeEditor/React/Model/Searcher/UndoRedo.hs
```diff
@@ -0,0 +1,74 @@
...
+
+-- === Public API === --
+
+mk :: InputState -> UndoRedoState
+mk = UndoRedoState def def
+
+undo :: UndoRedoState -> UndoRedoState
+undo st = case st ^. undoStack of
+    []         -> st
+    inp : inps -> st & undoStack .~ inps
```
iamrecursion: `inp` and `imps` aren't clear to me. What actually are they?

luna-studio/node-editor-view/src/NodeEditor/React/Event/Searcher.hs
kustosz: done

luna-studio/node-editor-view/src/NodeEditor/React/Model/Searcher/UndoRedo.hs
kustosz: Done by enabling strict for the file

luna-studio/node-editor-view/src/NodeEditor/React/Model/Searcher/UndoRedo.hs
kustosz: Changed to `input` and `inputs`.

luna-studio/node-editor-view/src/NodeEditor/React/Model/Searcher/UndoRedo.hs
```diff
@@ -0,0 +1,76 @@
...
+
+undo :: UndoRedoState -> UndoRedoState
+undo st = case st ^. undoStack of
+    []             -> st
+    input : inputs -> st & undoStack .~ inputs
+                         & redoStack %~ (input :)
+
+redo :: UndoRedoState -> UndoRedoState
+redo st = case st ^. redoStack of
+    []         -> st
```
wdanilo: alignment

PR #1354, opened 2019-03-27
Searcher snippets

kustosz:
### Pull Request Description

This adds the notion of "snippets" (ready to use expressions rather than just functions) to the searcher. The snippets are described in a config file in each library and searcher will find this file and display the snippets if found.

### Important Notes

@mwu-tow @sylwiabr there's a question for you. In unifying the structure of stdlib between luna and luna-studio (which is kind of an accidental change after bumping the luna version), I had to move the std sources from `env` to `env/stdlib`. Same should happen with Dataframes. What needs changing so that this structure works?

### Checklist


mwu-tow:
As for changing location where Dataframes are placed it should be just adjusting path here: https://github.com/luna/luna-manager/blob/8ce611a22bd99e7a07677767026834aec001d89f/luna-manager/src/Luna/Manager/Command/CreatePackage.hs#L238

libs/batch/plugins/luna-empire/src/Empire/Server/Graph.hs
```diff
@@ -107,8 +108,7 @@ import Empire.Empire                 (Empire)
 import Empire.Env                    (Env)
 import Empire.Server.Server          (defInverse, errorMessage, modifyGraph,
                                       modifyGraphOk, replyFail,
-                                      replyOk, replyResult, sendToBus',
-                                      webGUIHack)
+                                      replyOk, replyResult, sendToBus')
```
wdanilo: Why don't we use qualified names here? We should rather try to use as few direct imports as possible. Instead of `replyOk` we can use `Server.replyOk` etc – it's more clear and allows us to lower the maintenance burden.

libs/luna-empire/src/Empire/Commands/Graph/SearcherHints.hs
```diff
@@ -0,0 +1,156 @@
...
+snippetsFileName = "snippets.yaml"
+
+globalSnippetsFieldName :: Text
+globalSnippetsFieldName = "$global"
+
+importedSnippetsFieldName :: Text
+importedSnippetsFieldName = "$imported"
+
+getImportPaths ::
+    (MonadIO m, MonadException PackageNotFoundException m,
```
wdanilo: Use `Throws PackageNotFoundException m` instead of `MonadException`. We should remove `MonadException` from the public API.

libs/luna-empire/src/Empire/Commands/Graph/SearcherHints.hs
```diff
@@ -0,0 +1,156 @@
+module Empire.Commands.Graph.SearcherHints where
```
wdanilo: Very nicely layouted file! Can we just add the headers (`--=== ... === --`) to it in order to particion its contents according to semantics of the functions please?

libs/luna-empire/src/Empire/Commands/Graph/SearcherHints.hs
```diff
@@ -0,0 +1,156 @@
...
+getImportPaths ::
+    (MonadIO m, MonadException PackageNotFoundException m,
+     MonadException DatafileException m,
+     MonadException Path.PathException m
+     ) => GraphLocation -> m [FilePath]
+getImportPaths location = do
+    let file = location ^. GraphLocation.filePath
+    stdPath         <- StdLocator.findPath
+    filePath        <- Exception.rethrowFromIO @Path.PathException
+                                               (Path.parseAbsFile file)
```
wdanilo: This is a very minor one, and you can leave it as is, as we did not yet write down all the rules, but I believe this style of formatting keeps high mainenance burden, cause if you change the expression in line `66` you need to change the next line as well. Moreover, this style leaves realy a lot of unused white spaces. I'd definitely format it like:

```haskell

    stdPath         <- StdLocator.findPath
    filePath        <- Exception.rethrowFromIO @Path.PathException
                     $ Path.parseAbsFile file
```

libs/luna-empire/src/Empire/Commands/Graph/SearcherHints.hs
```diff
@@ -0,0 +1,156 @@
...
+getSnippetsFile projectRoot = do
+    contents <- liftIO $ Yaml.decodeFileEither (projectRoot <> snippetsFileName)
+    case contents of
+        Left err -> pure def
+        Right f  -> pure f
+
+isPublicMethod :: IR.Name -> Bool
+isPublicMethod (convert -> n) = head n /= Just '_'
+
+addSnippetsToLibrary :: RawLibrarySnippets -> SearcherLibrary.Library
```
wdanilo: Minor formatting (but lets just keep it in the new code), let's break the line on `::`, so we get

```haskell
addSnippetsToLibrary
    :: RawLibrarySnippets -> SearcherLibrary.Library -> Searcher.Library
```
instead please. Again, when changing the name of the function, we do not need to re-format all the lines + the signature starts with the same indentation with all the functions in top level, which makes it much easier to read.

libs/luna-empire/src/Empire/Commands/Graph/SearcherHints.hs
wdanilo: This file is not linted :(

luna-studio/node-editor-view/src/NodeEditor/React/Model/Searcher/Hint/Node.hs
```diff
@@ -132,11 +148,9 @@ instance Default Database where
 -- === API === --
 
 missingLibraries :: Getter Database (Set Library.Name)
-missingLibraries = to $ \d -> let
-    allHints         = Searcher.elems $ d ^. database
-    addLibName acc h = Set.insert (h ^. library . Library.name) acc
-    presentLibs      = foldl addLibName mempty allHints
-    importedLibs     = d ^. imported
+missingLibraries = to $ \db -> let
+    presentLibs  = Set.fromList $ Map.keys $ db ^. bareLibs
```
wdanilo: Let's try to lint the new code. I know that it's hard when we're changing only some part of old codebase, but such things like this line we should try to lint manually – everytime we see multiple `$`, we should refactor it to composition.

Issue #1323, opened 2019-01-12
Editor deletes part of text being written on 'breadcrumb' error


radeusgd:
### General Summary

When writing code in the editor, I think it tries to suggest variable names or something, but fails to do so and throws lots of errors.
Afterwards, parts of the text I was writing get deleted, so that I cannot really write code.

### Steps to Reproduce

1. Create a new project
2. Paste this code (other may work, but that's exactly what I've been working on)
```
import Std.Time
import Std.Base

class Looper a:
    Done a
    Next a (Looper a)

    def isStream: True

    def get: case self of
        Done a: a
        Next _ l: l.get

    def each f: case self of
        Done a: f a
        Next a l:
            f a
            l.each f

def unfoldLoop times initial f:
    case times of
        0: Done initial
        i:
            nxt = f initial
            Next initial (unfoldLoop times.pred nxt f)

def main:
    unfoldLoop1 = unfoldLoop 500 0 (i: seq (sleep 150.miliseconds.toMicroseconds) (i + 1))
    None
```
3. Try *manually* modifying main to
```
def main:
    unfoldLoop1 = unfoldLoop 500 0 (i: seq (sleep 150.miliseconds.toMicroseconds) (i + 1))
    unfoldLoop2 = unfoldLoop1.each (i: i * 100)
    None
```

### Expected Result

The code is modified.

### Actual Result

I get multiple red errors in the corner, and words/letters that I've written get randomly removed, so that I all managed to write was
```
def main:
    unfoldLoop1 = unfoldLoop 500 0 (i: seq (sleep 150.miliseconds.toMicroseconds) (i + 1))
    unfoldLoop
   None
```

The error I was getting (multiple times) has following log:
```
Breadcrumb {_items = [Definition {_nodeId = f22a1b23-2289-4363-b641-811422d07348}]}

while processing event

empire.atom.file.substitute.response

======= logs: =======

=== luna-empire-stdout---supervisor-U8fNCW.log ===
INFO 2019-01-12T13:13:44.174558596Z	:: received empire.graph.program.request
INFO empire.graph.program.request: Request {_location = GraphLocation {_filePath = "/home/radeusgd/luna/projects/LunaLooper/src/Main.luna", _breadcrumb = Breadcrumb {_items = []}}, _prevLocationSettings = Nothing, _retrieveLocation = True}
INFO 2019-01-12T13:13:44.242813907Z	:: sending response for empire.graph.program.request
[...]

PR #1356, opened 2019-04-04
Properly handle undo-redo responses in text

kustosz:
### Pull Request Description

This fixes handling updates from backend when using the text editor. Fixes the instances of #1323. Also a bunch of random reformattings in the touched files.

### Important Notes

The way we handle this is to handle Substitute updates selectively – when they arise when typing, they are ignored (since the Atom state already knows about tem). When they arise as a result of using undo-redo, they are applied, thus making the whole system behave properly and not apply states when it would disturb the user.

### Checklist

### DIFF
--- a/luna-studio/text-editor/src/TextEditor/Action/Batch.hs
+++ b/luna-studio/text-editor/src/TextEditor/Action/Batch.hs
@@ -1,14 +1,17 @@
 module TextEditor.Action.Batch  where
 
-import           Common.Action.Command         (Command)
-import           Common.Prelude
-import           Data.UUID.Types               (UUID)
-import           LunaStudio.Data.GraphLocation (GraphLocation)
-import           LunaStudio.Data.Range         (Range)
-import           LunaStudio.Data.TextDiff      (TextDiff)
-import           TextEditor.Action.UUID        (registerRequest)
-import qualified TextEditor.Batch.Commands     as BatchCmd
-import           TextEditor.State.Global       (State, clientId)
+import Common.Prelude
+
+import qualified TextEditor.Batch.Commands as BatchCmd
+import qualified TextEditor.State.Global   as State
+
+import Common.Action.Command         (Command)
+import Data.UUID.Types               (UUID)
+import LunaStudio.Data.GraphLocation (GraphLocation)
+import LunaStudio.Data.Range         (Range)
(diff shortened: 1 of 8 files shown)


luna-studio/text-editor/src/TextEditor/Action/Batch.hs
```diff
@@ -45,7 +48,11 @@ setProject :: FilePath -> Command State ()
 setProject = withUUID . BatchCmd.setProject
 
 substitute :: GraphLocation -> [TextDiff] -> Command State ()
-substitute = withUUID .: BatchCmd.substitute
+substitute location diffs = do
+    uuid  <- registerRequest
+    guiId <- use clientId
+    State.ignoreResponse uuid
+    liftIO $ BatchCmd.substitute location diffs uuid (Just guiId)
```
wdanilo: If we are adding `liftIO` its better to add it on the call side - it will not break existing code and will conform to the rules

luna-studio/text-editor/src/TextEditor/Handler/Text.hs
```diff
@@ -2,48 +2,64 @@ module TextEditor.Handler.Text
...
+handle (Batch (CopyResponse  response))
+    = Just $ handleResponse response success doNothing2 where
+        success result = do
+            let uri  = response ^. Response.request . Copy.filePath
+                code = result ^. Copy.code
+            liftIO $ JS.setClipboard (convert uri) (convert code)
 
 handle (Batch (SubstituteUpdate (Substitute.Update path diffs))) =
-    Just $ liftIO $ JS.insertCode $ TextEvent (GraphLocation path def) diffs
+    Just . liftIO . JS.insertCode $ TextEvent (GraphLocation path def) diffs
```
wdanilo: This liftIO does not make sense here. I think we should create automatic check for unnecessary liftIOs to discover them by the compiler. Something like

```haskell
class LiftIO m where
  liftIO :: MonadIO n => m a -> n a
instance LiftIO IO where
  liftIO = Old.liftIO
instance {-# OVERLAPPABLE #-} Error "Unnecessary liftIO usage" => LiftIO m where 
  liftIO = impossible
```

Writing this to gather comments @kustosz @iamrecursion

PR #1361, opened 2019-04-09
Search by tags

kustosz:
### Pull Request Description
This adds searching by function tags to the node searcher.
Since there isn't such a thing as "tags" yet, it works on a mocked structure, adding a bunch of tags for arithmetic operators. This is to be replaced by a proper way of parsing them from the documentation.

### Important Notes
Nothing interesting happens here, very routine changes.

### Checklist

### DIFF
--- a/libs/luna-studio-common/src/LunaStudio/Data/Searcher/Hint.hs
+++ b/libs/luna-studio-common/src/LunaStudio/Data/Searcher/Hint.hs
@@ -21,6 +21,7 @@ import Data.Text    (Text)
 data Raw = Raw
     { _name          :: Text
     , _documentation :: Text
+    , _tags          :: [Text]
     } deriving (Eq, Generic, Show)
 
 makeLenses ''Raw
@@ -33,4 +34,8 @@ instance ToJSON Raw where
     toEncoding = LensAeson.toEncoding
 
 instance FromJSON Raw where
-    parseJSON = LensAeson.parse
+    parseJSON = Aeson.withObject "Raw" $ \o -> do
+        name <- o Aeson..:  "name"
+        doc  <- o Aeson..:  "documentation"
+        tags <- o Aeson..:? "tags" Aeson..!= mempty
+        pure $ Raw name doc tags
--- a/luna-studio/node-editor-view/src/NodeEditor/React/Model/Searcher/Hint/Node.hs
+++ b/luna-studio/node-editor-view/src/NodeEditor/React/Model/Searcher/Hint/Node.hs
@@ -55,13 +55,15 @@ data Node = Node
     , _library       :: Library.Info
     , _kind          :: Kind
     , _documentation :: Text
+    , _tags          :: [Text]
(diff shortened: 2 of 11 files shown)


libs/luna-studio-common/src/LunaStudio/Data/Searcher/Hint.hs
```diff
@@ -33,4 +34,8 @@ instance ToJSON Raw where
     toEncoding = LensAeson.toEncoding
 
 instance FromJSON Raw where
-    parseJSON = LensAeson.parse
+    parseJSON = Aeson.withObject "Raw" $ \o -> do
+        name <- o Aeson..:  "name"
+        doc  <- o Aeson..:  "documentation"
+        tags <- o Aeson..:? "tags" Aeson..!= mempty
+        pure $ Raw name doc tags
```
wdanilo: Wy do we use manual implementation here? Does the LensAeson.parse do bad job here?

luna-studio/node-editor-view/src/NodeEditor/React/Model/Searcher/Hint/Node.hs
```diff
@@ -162,7 +165,7 @@ mkLocalFunctionsDb :: [Text] -> Database
 mkLocalFunctionsDb syms = insertSearcherLibraries libs def where
     libs    = Map.singleton localFunctionsLibraryName library
     library = Library.Library hints def def def
-    hints   = flip Hint.Raw mempty <$> syms
+    hints   = (\sym -> Hint.Raw sym mempty mempty) <$> syms
```
wdanilo: I'd consider using a smart constructor here – there are other places where we use `Raw <something> mempty mempty`. Using smart cons here could help with readability and will simplify the code as well.

luna-studio/node-editor-view/src/NodeEditor/React/View/Searcher.hs
```diff
@@ -126,22 +126,30 @@ results_ ref selected wait results = when (not (null results) || wait) $ do
...
+                                    $ result ^. SearcherData.text
+
+
+highlighted_ :: Result Hint -> ReactElementM ViewEventHandler ()
+highlighted_ result = go 0 highlights where
+    highlights = result ^. Result.match . Match.range
+    name       = convert $ result ^. SearcherData.text
+    go :: Int -> [Range] -> ReactElementM ViewEventHandler ()
+    go omit [] = span_ [ "key" $= "l" ] . elemString
+                                        $ drop omit name
```
wdanilo: `$` indentation should be equals to `=` indentation

libs/luna-studio-common/src/LunaStudio/Data/Searcher/Hint.hs
kustosz: It's about the `tags` field. Aeson does not allow this field to be missing by default (same for LensAeson), but I don't want everyone typing snippet files in yaml to have to include an empty tags list for each entry.

PR #1364, opened 2019-04-12
Base snippet

sylwiabr:
### Pull Request Description

Improve the code snippets for standard library
### Checklist


env/stdlib/Std/snippets.yaml
```diff
@@ -1,5 +1,207 @@
-Std.HTTP:
+Std.Base:
   $global:
+    - name: pi
+      documentation: The mathematical constant π = 3.14159265359.
+    - name: Map.singleton "foo" 10
+      documentation: Creates a map with a single element.
```
iamrecursion: I would say what the key and value are. Just to make it more explicit.

env/stdlib/Std/snippets.yaml
```diff
@@ -1,5 +1,207 @@
...
+    - name: JSON.empty
+      documentation: Empty JSON object.
+    - name: readFile "/example/path/to/data"
```
iamrecursion: Can we give them a snippet to a known good file instead? I'm not sure what the point of one is that is just a dummy path.

env/stdlib/Std/snippets.yaml
```diff
@@ -1,5 +1,207 @@
...
+      documentation: Read file from the hard drive.
+    - name: writeFile "/example/path/to/file" "lorem ipsum"
```
iamrecursion: Like above, a propely writeable file might be better, especially given these are likely going to execute on load into LS.

env/stdlib/Std/snippets.yaml
```diff
@@ -1,5 +1,207 @@
...
+      documentation: TimeInterval in minutes.
+    - name: hours
+      documentation: TimeInterval in hours.
+    - name: days
+      documentation: TimeInterval in days.
+    - name: weeks
+      documentation: TimeInterval in weeks.
+  Real:
+    - name: + 3.0
+      documentation: Add two floats.
```
iamrecursion: We call it `Real`, so do we want the snippets to refer to them as 'floats'?

env/stdlib/Std/snippets.yaml
```diff
@@ -1,5 +1,207 @@
...
+      documentation: Return list of all characters in the text.
+    - name: lowercase
+      documentation: Change all characters to lowercase.
+    - name: reverse
+      documentation: Reverse the text.
+    - name: uppercase
+      documentation: Change all characters to uppercase.
+  Complex:
+    - name: modulus
+      documentation: Calculate modulus.
```
iamrecursion:
```suggestion
      documentation: Calculate complex modulus.
```

While this is clear in many contexts, there are some where it wouldn't be without the addition.

env/stdlib/Std/snippets.yaml
```diff
@@ -1,5 +1,207 @@
...
+  Stream:
+    - name: drop 1
+      documentation: Drop the given number of elements from the stream.
+    - name: take 1
+      documentation: Return the prefix of the stream of given length.
+    - name: consume 5
+      documentation: List of the first `n` elements of stream and the result of dropping the first `n` values of it.
+
+Std.Geo:
+  GeoJSONFeature:
```
iamrecursion: Would be nice to have a few more for this.

env/stdlib/Std/snippets.yaml
```diff
@@ -1,5 +1,207 @@
...
+    - name: addProperty "popupContent" "home"
+      documentation: Add a property do GeoJSONFeature.
+
+Std.HTTP:
+  $global:
```
iamrecursion: And here too.

env/stdlib/Std/snippets.yaml
sylwiabr: We will as soon as https://github.com/luna/luna/pull/394 will be merged

env/stdlib/Std/snippets.yaml
sylwiabr: then we need to improve the library :)

env/stdlib/Std/snippets.yaml
iamrecursion: Brilliant. If you want to merge this ASAP, then maybe file an issue to track that improvement!

sylwiabr: https://github.com/luna/luna-studio/issues/1366

sylwiabr:
It will come to Luna studio with Luna bump

PR #1368, opened 2019-04-25
Refactor project structure

kustosz:
### Pull Request Description
There was a bunch of very non-sensical project structure choices, inconsistent naming schemes etc. This PR aims to resolve it.

### Important Notes
* We had backend files scattered throughout 2 different directories, with very unclear boundaries and some old old old legacy naming. They all reside inside `backend` dir now.
* The name `luna-studio` for the frontend directory was very confusing (it's _all_ Luna Studio, why would it a part of the project be named the same?). It's `frontend` now.
* The name `luna-empire` is bad. The binary for this is now called `luna-project-manager`. The name `luna-empire` still persists in the library part (to be renamed when I get to refactoring the whole of it). The name `Empire` also still appears in the source code for `luna-project-manager`. Cleaning up that whole module is next on my plate, so it will disappear soon.
* Some of the names and paths were hardcoded in multiple places (build scripts, supervisor configs, windows service configs). I did my best to fix it. It works on my machine. I may have broken something anyway, we need to carefully test packages with this.

### Checklist


kustosz:
Oh also some dead code removal.

wdanilo:
Regarding @mwu-tow comment, if the "project manager" is not managing user projects, it might be an unfortunate name.

kustosz:
OK, so:
ad @mwu-tow's 2. The name is unfortunate and responsibilities are ill-defined. This service does 3 things:
1. Project management (to be split out)
2. Double representation management (bulk of its operations)
3. Typechecking / interpretation (to be split out)
Thus, after consultations with @wdanilo: All services will be under `services` directory. I'm splitting out project management soon (not in this PR though) and reserving `project-manager` for the new service. This one will be named `double-representation`. Interpreter will be split out after it becomes viable to serialize Luna IR's for inter-process communication.

ad @iamrecursion 2. and @mwu-tow 1. – We can't do both :P which one do we settle on? I'm with @mwu-tow on this one, just to minimize the number of not-really-doing-anything-interesting directories, but it's not a strong preference.

ad @iamrecursion 1. I disagree – `luna-studio-common` is not a library in a sense of being in any way generic. It's an API definition library, thus a `common` module between frontend and backend. Also it probably should be `common/api-definition` rather than `common/luna-studio-common`.

iamrecursion:
> ad @iamrecursion 2. and @mwu-tow 1. – We can't do both :P

If we don't want to keep `build_config`, then yes we definitely should put it where @mwu-tow said.

As for the rename, fine! I like the suggestion of `common/api-definition`

mwu-tow:
@kustosz Well, obviously i support the solution I proposed. :-) Arguments:
1) consistency with frontend/runner — we already have this scheme adopted for other parts, it should be all or nothing.
2) `luna` repository is a different beast — it provides a single `stack.yaml` in root as it is esentially a single "project" (in meaning of "set of buildable targets, as in stack.yaml). `luna-studio` repository is essentially three project — this is not a "config", this is just the project file for the targets in the subdirectories.

wdanilo:
@sylwiabr is late to the party :P

kustosz:
Oh cool, I got the approves before fixing stuff! Disregard my previous comment, merging as-is 😈

PR #1372, opened 2019-05-08
Communications refactor

kustosz:
### Pull Request Description
This purges the oldest and most arcane part of the project – the ZMQ based communication framework. It is a great simplification of this whole thing – the old library "provided" a lot of functionality that was never actually useful for Luna (like a poor-man's RPC module and an unnecessary complication in the network topology).
This PR removes that old codebase and instead introduces a very simple and small ZMQ wrapper, providing only all-to-all communications (which is all Luna uses and probably will use). Even if we ever wanted to specify some other topology, the old library was equally ill-suited for that (i.e. both would need coding that pattern from scratch).

### Important Notes
Keep in mind this PR touches a lot of files inside `double-representation` and `undo-redo` modules. They are ugly and their refactoring should happen further down the line. This PR concerns the new ZMQ system.

### Checklist


backend/libs/bus/src/Bus/Data/Message.hs
```diff
@@ -0,0 +1,55 @@
...
+import qualified Data.Text.Encoding as Encoding
+import qualified Data.Text as Text
+
+import Data.Binary (Put, Get)
+import Data.ByteString (ByteString)
+
+
+-- === Definition === --
+
+type Topic = String
```
iamrecursion: Based on our discussion, I'd like to see a comment explaining why this is left stringly typed. The reasoning is clear and entirely valid, it just needs to be _written down_!

backend/libs/bus/src/Bus/Framework/App.hs
```diff
@@ -0,0 +1,91 @@
+{-# LANGUAGE UndecidableInstances #-}
```
iamrecursion: What's this used for?

backend/services/double-representation/src/Empire/Server.hs
```diff
@@ -7,98 +7,87 @@
...
-import qualified Luna.Pass.Resolve.Data.Resolution  as Res
-import qualified Luna.Pass.Scheduler                 as Scheduler
-import qualified Luna.Pass.Sourcing.UnitMapper as UnitMap
-import qualified Luna.Std as Std
-import qualified Data.Bimap as Bimap
-import qualified Path
-import qualified Luna.Pass.Flow.ProcessUnits as ProcessUnits
+import Bus.Data.Config               (Config)
+import Bus.Data.Message              (Message, Topic)
+import Control.Concurrent            (forkIO, forkOn)
```
iamrecursion: These should probably be used qualified. There are a few other imports that also should be below. Rule of thumb is to only import operators and the 'primary type' (e.g. `TChan` for `STM.TChan`) unqualified.

backend/services/double-representation/src/Empire/Server/Library.hs
```diff
@@ -3,32 +3,33 @@
...
 import qualified System.Log.MLogger                   as Logger
-import           ZMQ.Bus.Trans                        (BusT (..))
+
+import Control.Lens           (to, use, (.=), (^..))
+import Control.Monad.Catch    (try)
+import Control.Monad.State    (StateT)
+import Empire.Commands.Graph  (prepareLunaError)
+import Empire.Data.AST        (SomeASTException)
+import Empire.Env             (Env)
+import Empire.Server.Server   (replyFail, replyResult, sendToBus')
```
iamrecursion: These should be imported qualified.

backend/services/double-representation/src/Empire/Server/Server.hs
```diff
@@ -1,80 +1,77 @@
...
-import           LunaStudio.Data.GraphLocation (GraphLocation (..))
 import qualified LunaStudio.Data.GraphLocation as GraphLocation
 import qualified LunaStudio.Data.Node          as Node
 import qualified System.Log.MLogger            as Logger
-import qualified ZMQ.Bus.Config                as Config
-import qualified ZMQ.Bus.Data.Message          as Message
-import qualified ZMQ.Bus.EndPoint              as EP
-import           ZMQ.Bus.Trans                 (BusT (..))
 
+import Control.Arrow                 ((&&&))
```
iamrecursion: A bunch more things here that should be imported qualified.

backend/services/undo-redo/src/Undo.hs
```diff
@@ -5,37 +5,32 @@
...
-import           ZMQ.Bus.EndPoint          (BusEndPoints)
-import qualified ZMQ.Bus.Trans             as Bus
-
-import qualified System.IO as IO
-
 
+import Bus.Data.Config      (Config)
+import Bus.Data.Message     (Message)
+import Control.Lens         (uses, (%=))
+import Control.Monad.State  (MonadState, StateT (StateT), evalStateT, runStateT)
```
iamrecursion: A few more things to be imported qualified.

backend/services/double-representation/src/Empire/Server.hs
kustosz: Correct. This whole file is a steaming pile of... mess. Needs a deeper rewrite, wontfix for now.

backend/services/double-representation/src/Empire/Server/Library.hs
kustosz: Same, file needs rewrite

backend/services/double-representation/src/Empire/Server/Server.hs
kustosz: Same, file needs rewrite

backend/services/undo-redo/src/Undo.hs
kustosz: Same – deeper refactor needed later.

backend/libs/bus/src/Bus/Framework/App.hs
kustosz: Confusing the readers! Removed ;)

PR #1376, opened 2019-05-09
Replace shelly with typed-process

### LINKED ISSUE #1374: Remove Shelly Dependency from Luna Studio Runner
### Summary
Currently, a very old `shelly` dependency is keeping [`luna-studio-runner`](https://github.com/luna/luna-studio/blob/master/runner) on a very old LTS (`lts-9.2`). As part of the efforts to modernise this code-base to a newer LTS, it needs to be re-written to use the standard `process` API, rather than `shelly`. 

### Value
Doing this has twofold benefits:

- The package can depend on newer versions of our internal libraries, used directly from the Luna tree.
- The package is closer to allowing Luna Studio to be packaged using `nix` as discussed in https://github.com/luna/luna-manager/issues/127. 

### Specification

- Remove the dependency on the `shelly` library.
- Re-write `luna-studio-runner` to make use of the standard `process` API.

### Acceptance Criteria & Test Cases

- `luna-studio-runner` builds without a dependency on `shelly`. 
- It still performs all of its duties correctly in a Luna Studio package.

mulderr:
### Pull Request Description
Attempts to resolve #1374.

### Important Notes
- replace `shelly` with `typed-process` and `path-io`
- some general cleanup
- ensure temporary directories are removed

### Checklist

- [x] The code conforms to the [Luna Style Guide](https://github.com/luna/luna/blob/master/doc/haskell-style-guide.md) if it is Haskell.


runner/src/StudioRunner.hs
```diff
@@ -240,41 +239,48 @@ unixOnly act = case currentHost of
...
     packageAtomHome <- packageStudioAtomHome
-    atomHomeParent  <- parent <$> userStudioAtomHome
-    Shelly.shelly $ do
-        Shelly.mkdir_p atomHomeParent
-        Shelly.cp_r packageAtomHome atomHomeParent
-    when (currentHost == Windows) $ liftIO $ runProcess_ $ shell $ "attrib +h " <> (encodeString mainHomePath)
+    atomHomeParent  <- takeDirectory <$> userStudioAtomHome
+    liftIO $ do
+      createDirectoryIfMissing True atomHomeParent
+      -- TODO: find a library that can handle recursive copies while preserving symlinks on unix
```
mwu-tow: I don't know such library, however I don't think we care about symlinks here. Are there any symlinks in the ~/.luna contents?
If we don't require symlinks, then Cabal's [`copyDirectoryRecursive`](https://www.haskell.org/cabal/release/cabal-latest/doc/API/Cabal/Distribution-Simple-Utils.html#v:copyDirectoryRecursive) can be used.

runner/src/StudioRunner.hs
```diff
@@ -240,41 +239,48 @@ unixOnly act = case currentHost of
...
+    versionN  <- T.strip <$> versionText
+    resources <- resourcesDirectory
+    localShareFolder <- sharePath
+    let iconsFolder  = resources </> "icons"
+        desktopFile  = resources </> "app_shared.desktop"
+        localDesktop = localShareFolder </> "applications" </> T.unpack (T.concat ["LunaStudio", versionN, ".desktop"])
+    liftIO $ do
+      createDirectoryIfMissing True $ takeDirectory localShareFolder
+      createDirectoryIfMissing True $ takeDirectory localDesktop
+      -- TODO: find a library that can handle recursive copies while preserving symlinks on unix
```
mwu-tow: There's none, AFAIK. And I don't really think that there is anything wrong with using `cp -r` in platform-specific branch. Perhaps we should just refactor the cp process call to a separate function named like copyDirectoryRecursiveUnix and that's it. I'd rather have us rely on well-tested program (like `cp`) instead of trying to implement it from scratch and fail (like `shelly`).

runner/src/StudioRunner.hs
```diff
@@ -287,19 +293,22 @@ checkLunaHome = do
...
     supervisorBinPath <- supervisorctlBinPath
-    supervisorDir     <- backendDir
-    let runSupervisorctl  = Shelly.chdir supervisorDir $ Shelly.run supervisorBinPath args
-        supressErrors act = Shelly.catchany act (\_ -> return "Unable to run supervisorctl")
-    liftIO . Shelly.shelly $ supressErrors runSupervisorctl
+    supervisorDir <- backendDir
+    let runSupervisorctl = readProcess_ $ setWorkingDir supervisorDir
+                                        $ proc supervisorBinPath $ map T.unpack args
+        supressErrors act = do
+          a <- Async.async act
```
mwu-tow: Why do we need `async`? It doesn't seem equivalent ot the original code.

mwu-tow:
As for the code style guide it has been just moved here: https://github.com/luna/luna/blob/master/doc/haskell-style-guide.md
@iamrecursion I believe our issue/pr templates needs updating then.

runner/src/StudioRunner.hs
```diff
@@ -240,41 +239,48 @@ unixOnly act = case currentHost of
...
-        Shelly.mkdir_p atomHomeParent
-        Shelly.cp_r packageAtomHome atomHomeParent
-    when (currentHost == Windows) $ liftIO $ runProcess_ $ shell $ "attrib +h " <> (encodeString mainHomePath)
+    atomHomeParent  <- takeDirectory <$> userStudioAtomHome
+    liftIO $ do
+      createDirectoryIfMissing True atomHomeParent
+      -- TODO: find a library that can handle recursive copies while preserving symlinks on unix
+      if currentHost == Windows
+      then do
+        runProcess_ $ proc "robocopy" [packageAtomHome, atomHomeParent, "/e"]
```
mwu-tow: Shouldn't we manually check exit code, as robocopy doesn't stick to 0-on-success?
See https://docs.microsoft.com/en-us/windows-server/administration/windows-commands/robocopy#exit-return-codes

runner/src/StudioRunner.hs
mulderr: Oh shoot, I'm in trouble. I seem to recall reading that catching SomeException is almost like an anti-pattern because... mumble, mumble... asynchronous exceptions... mumble, mumble... best just use `async`. I consider not understanding async exceptions one of my cardinal sins when writing Haskell code :)

Can we stick to catching SomeException like shelly does internally? If so then we can remove `async` altogether.

runner/src/StudioRunner.hs
mulderr: Good catch. I'll wait with fixing this for now because if it turns out we can ignore symlinks on linux than both of the `runProcess_` calls in `copyLunaStudio` could be replaced by `copyDirectoryRecursive` or so.

iamrecursion:
> @iamrecursion I believe our issue/pr templates needs updating then.

Indeed, but that guide is very out of date. It’s on my list for today to update the style guide itself and then update the links in the PR templates.

runner/src/StudioRunner.hs
mulderr: Yep that's always a reasonable fallback but if we forgo the symlink requirement then a straightforward recursive copy seems very doable in Haskell. We wouldn't have to rely on things in PATH which I guess is always good security wise.

iamrecursion:
I have now updated the style guide, which can be found at the link posted above. Future PRs will have the correct link in their template.

runner/src/StudioRunner.hs
mwu-tow: Oh, I'm a C++ programmer and I feel you so much when it comes to dealing with exceptions in Haskell. From what I understand asynchronous exceptions are cross-thread, so I don't think it is something we have to deal in the runner. 

We have previously catched SomeException through `Shelly.catchany`, I think that switching to `Control.Exception.Safe` equivalent [`catchAny`](https://www.stackage.org/haddock/lts-13.13/safe-exceptions-0.1.7.0/Control-Exception-Safe.html#v:catchAny) or sth similar for now is fine — the point od this PR is to get rid of shelly, not rewrite exception handling.
We do have our own [library for exceptions](https://github.com/luna/luna/tree/master/lib/exception) that should be in introduced in the long term… but that's not for this PR

mulderr:
I'm using a debian 9 vm. Did `luna-manager develop luna-studio` to set things up and then just replaced `~/luna-develop/apps/luna-studio` with `git clone` of my fork. Not sure what is the standard way to build a custom branch/commit. Results pending.

For no reason whatsoever I'm going to also remark that nix supports custom binary caches in addition to cache.nixos.org that usually already has most of hackage. Again. No reason.

mwu-tow:
> I'm using a debian 9 vm. Did luna-manager develop luna-studio to set things up and then just replaced ~/luna-develop/apps/luna-studio with git clone of my fork. Not sure what is the standard way to build a custom branch/commit. Results pending.

To create the package, use luna-manager's `make-package` command, giving a version number (might be arbitrary) and path to built luna-studio's `luna-package.yaml` file. It doesn't necessarily have to be in ~/luna-develop, though this the default setup and should work.

However, since you only touch runner that is self-contained, single executable, you should be to just build it and replace it in the pre-built luna-studio package. This is also a little troublesome, as you will need to edit our AppImage file… but shouldn't be that hard. (or just extracting / copying it and patching should work as well I believe)

mulderr:
@mwu-tow Yeah, my improvised way in the end fails with a python exception `DistutilsFileError: cannot copy tree ...`. Trying again with `luna-manager make-package <path-to-fork> 1.1.1`. When giving an arbitrary version number it complained about unknown version. Had to give it the exact version from `luna-package.yaml` and now it seems to be building the backend.

Do I guess correctly that `make-package` is going to use the release config so no `-fomit-interface-pragmas`? If that is the case then it's likely I will have to edit the AppImage anyway but let's try the standard route first.

mulderr:
Build on linux ok. luna-studio studio runs ok. On first start it complains about missing `user_info.json`. Subsequent starts ok. Investigating before moving to windows.

mwu-tow:
> When giving an arbitrary version number it complained about unknown version. Had to give it the exact version from luna-package.yaml and now it seems to be building the backend.

That's strange, I have built really stupid version numbers with no issue…
Well, you got it building and that's most important.

> Do I guess correctly that make-package is going to use the release config so no -fomit-interface-pragmas? If that is the case then it's likely I will have to edit the AppImage anyway but let's try the standard route first.

Yes, if needed the flags passed can be patched here: https://github.com/luna/luna-studio/blob/master/scripts_build/stack_build.py#L48

> Moving on to windows.

Here I even stronger suggest just building the runner, setting up build environment for the proper package… is troublesome.
If you need, I can provide the binary — I'm primarily based on Windows, so have environment set up. Just let me know.

mulderr:
I am having problems making anything work on windows indeed. I would not mind that binary at all :)

Getting back to recursive copy, I do not see any symlinks anywhere. [copyDirRecur](http://hackage.haskell.org/package/path-io-1.4.2/docs/Path-IO.html#v:copyDirRecur) from `path-io` looks like a good choice. Maybe even go all in and replace all `FilePath`s with respective `Path b t` eventually. Both `path` and `path-io` look really nice.

runner/src/StudioRunner.hs
mulderr: Done. Removed async in favor of `catchAny`. Sorry for the noise :)

runner/src/StudioRunner.hs
mulderr: Using `copyDirRecur` so the problem went away.

runner/src/StudioRunner.hs
mulderr: Went with `copyDirRecur`. Hope that's fine.

runner/src/StudioRunner.hs
mulderr: Went with `copyDirRecur`. Hope that's fine.

mulderr:
Bar any symlink concerns or windows specific issues I think it should be ready to merge. Let me know if you'd prefer a different take on this.

runner/src/StudioRunner.hs
```diff
@@ -82,224 +78,216 @@ type MonadRun m = (MonadState RunnerConfig m, MonadIO m, MonadMask m)
...
+absHomeDir ::  MonadRun m => m (Path Abs Dir)
+absHomeDir = do
   runnerCfg <- get @RunnerConfig
-  home      <- decodeString <$> (liftIO getHomeDirectory)
+  home      <- getHomeDir
   return $ home </> (runnerCfg ^. mainHomeDir)
 
-relativeToDir :: MonadRun m => m FilePath -> [Getting FilePath RunnerConfig FilePath] -> m FilePath
-relativeToDir basePath segmentAccessors = do
+relToDirDir :: MonadRun m
```
mwu-tow: This name feels misleading to me — not only I have no idea what the function does, it is named as ig it were returning a relative path.
Perhaps something more like `buildDirectoryPath` or something?

runner/src/StudioRunner.hs
```diff
@@ -313,94 +301,84 @@ testIfRunning = do
...
-        setEnv "LUNA_VERSION_PATH"     =<< versionFilePath
+        setEnvPath "LUNA_STUDIO_DATA_PATH" =<< dataStorageDirectory develop
+        setEnvPath "LUNA_STUDIO_LOG_PATH"  =<< windowsLogsDir       develop
+        setEnvPath "ATOM_HOME"             =<< userStudioAtomHome
+        setEnvPath "LUNA_PROJECTS"         =<< lunaProjectsPath
+        setEnvPath "LUNA_USER_INFO"        =<< userInfoPath
+        setEnvPath "LUNA_VERSION_PATH"     =<< versionFilePath
         createStorageDataDirectory develop
-        bracket_ startServices stopServices $ Shelly.shelly $ Shelly.cmd atom
+        bracket_ startServices stopServices $ runProcess_ $ shell atom
```
mwu-tow: This fails when `atom` path contains spaces and such is the case with default installation path on Windows.
Generally, whenever feasible we should prefer `proc` over `shell` — it takes care of proper arguments quoting for us and prevents any other unexpected shell itnerferences.

runner/src/StudioRunner.hs
```diff
@@ -82,224 +78,216 @@ type MonadRun m = (MonadState RunnerConfig m, MonadIO m, MonadMask m)
...
-    return $ foldl (</>) base pathSegments
-
-relativeToMainDir, relativeToHomeDir :: MonadRun m => [Getting FilePath RunnerConfig FilePath] -> m FilePath
-relativeToMainDir = relativeToDir mainAppDir
-relativeToHomeDir = relativeToDir mainHomeFolder
+    b <- basePath
+    let segments = map (runnerCfg ^.) segmentAccessors
+    pure $ foldl (</>) b segments
+
+relToDirFile :: MonadRun m
```
mwu-tow: as above

runner/src/StudioRunner.hs
```diff
@@ -409,36 +387,47 @@ runPackage develop forceRun = case currentHost of
 
 runApp :: MonadRun m => Bool -> Bool -> Maybe String -> m ()
 runApp develop forceRun atom = do
-    liftIO $ Environment.setEnv "LUNA_STUDIO_ATOM_ARG" (fromMaybe " " atom)
+    setEnv "LUNA_STUDIO_ATOM_ARG" (fromMaybe " " atom)
     runPackage develop forceRun
 
+withLunaTempDir :: MonadRun m => m () -> m ()
```
mwu-tow: Were we previously deleting the tmp dir? Is this a behavior change?
(the change though does make much sense)

runner/src/StudioRunner.hs
```diff
@@ -12,68 +12,64 @@
...
+                                 , _supervisordConfig      :: Path Rel File
+                                 , _atomFolder             :: Path Rel Dir
+                                 , _thirdPartyFolder       :: Path Rel Dir
+                                 , _backendBinsFolder      :: Path Rel Dir
+                                 , _binsFolder             :: Path Rel Dir
+                                 , _packageFolder          :: Path Rel Dir
+                                 , _supervisorKillFolder   :: Path Rel Dir
+                                 , _supervisorKillBin      :: Path Rel File
+                                 , _atomBinPath            :: Path Rel File
+                                 , _mainTmpDirectoryTempl  :: String
```
iamrecursion: What's the reasoning behind this one being a string? It would be good to have a comment explaining the reasoning.

runner/src/StudioRunner.hs
```diff
@@ -82,224 +78,216 @@ type MonadRun m = (MonadState RunnerConfig m, MonadIO m, MonadMask m)
...
+        , _supervisorctlBin       = $(mkRelFile "supervisorctl")
+        , _supervisordConfig      = $(mkRelFile "supervisord-package.conf")
+        , _atomFolder             = $(mkRelDir "atom")
+        , _thirdPartyFolder       = $(mkRelDir "third-party")
+        , _backendBinsFolder      = $(mkRelDir "private")
+        , _binsFolder             = $(mkRelDir "bin")
+        , _packageFolder          = $(mkRelDir "packages")
+        , _supervisorKillFolder   = $(mkRelDir "kill")
+        , _supervisorKillBin      = $(mkRelFile "kill")
+        , _atomBinPath            = $(mkRelFile $ "atom" FP.</> "usr" FP.</> "bin" FP.</> "atom")
```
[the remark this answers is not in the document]
iamrecursion: They are. We use paths with `/` separators internally in the Luna codebase and have no issues on windows. From the documentation of `path`:

> A path is represented by a number of path components separated by a path separator which is a / on POSIX systems and can be a / or \ on Windows.

runner/src/StudioRunner.hs
```diff
@@ -82,224 +78,216 @@ type MonadRun m = (MonadState RunnerConfig m, MonadIO m, MonadMask m)
...
 instance Monad m => MonadHostConfig RunnerConfig 'Windows arch m where
     defaultHostConfig = reconfig <$> defaultHostConfigFor @Linux where
-        reconfig cfg = cfg & atomBinPath .~ ("Atom" </> "atom.exe")
+        reconfig cfg = cfg & atomBinPath .~ $(mkRelFile $ "Atom" FP.</> "atom.exe")
 
 -- path helpers --
-runnerDir :: MonadIO m => m FilePath
-runnerDir = (directory . decodeString) <$> liftIO getExecutablePath
+runnerDir :: MonadIO m => m (Path Abs Dir)
+runnerDir = liftIO $ do
```
iamrecursion: Please indent with four spaces!

runner/src/StudioRunner.hs
```diff
@@ -82,224 +78,216 @@ type MonadRun m = (MonadState RunnerConfig m, MonadIO m, MonadMask m)
...
+    let stepUp = if dirname runnerDir == $(mkRelDir "main")
+                 then parent . parent          -- drop bin/main/
+                 else parent . parent . parent -- drop bin/public/luna-studio/
     pure $ stepUp runnerDir
 
-mainHomeFolder ::  MonadRun m => m FilePath
-mainHomeFolder = do
+absHomeDir ::  MonadRun m => m (Path Abs Dir)
+absHomeDir = do
   runnerCfg <- get @RunnerConfig
```
iamrecursion: Identation.

runner/src/StudioRunner.hs
mulderr: It's template for a file name and not an actual file name. If you set this to `luna` then the temporary directory could be something like `/tmp/luna2718`. See [withSystemTempDirectory](http://hackage.haskell.org/package/temporary-1.3/docs/System-IO-Temp.html#v:withSystemTempDirectory). I'll add a comment to make this clear.

iamrecursion: Might be worth making it `mainTmpDirectoryTemplate` as well, just for clarity's sake!

runner/src/StudioRunner.hs
mulderr: It is a behavior change. Probably should have made this more clear.

If you look at [`lunaTmpPath`](https://github.com/luna/luna-studio/blob/6108368400ba06ff00a8986d70680876ecd201fd/runner/src/StudioRunner.hs#L206) it is unlike the rest of the path functions. It doesn't just build a path it creates the directory too. As far as I can tell no cleanup was ever done later. On my test vm I noticed a bunch of `/tmp/luna*` directories from previous runs. Additionally [`lunaTutorialsPath`](https://github.com/luna/luna-studio/blob/6108368400ba06ff00a8986d70680876ecd201fd/runner/src/StudioRunner.hs#L215) was not aware of this side effect of `lunaTmpPath` and that created yet another tmp directory that also was never cleaned anywhere.

Old behavior:
* two tmp directories (I assume this is accidental)
* no cleanup

New behavior:
* only one tmp directory
* tmp directory removed after luna-studio exits

runner/src/StudioRunner.hs
mulderr: Sure. That works.

runner/src/StudioRunner.hs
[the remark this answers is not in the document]
mulderr: Nice. This also means filepath is not needed and can be removed.

runner/src/StudioRunner.hs
```diff
@@ -12,68 +12,64 @@
...
 import           Control.Monad
 import           Control.Monad.IO.Class        (MonadIO, liftIO)
 import           Control.Monad.State.Lazy
-import           Data.ByteString.Lazy          (unpack)
-import           Data.List.Split
+import qualified Data.ByteString.Lazy          as BL
 import qualified Data.List                     as List
 import           Data.Maybe                    (fromMaybe, maybeToList)
 import           Data.Semigroup                ((<>))
 import qualified "containers"  Data.Set        as Set
```
mulderr: Do we need to use PackageImports here?

iamrecursion: Does it compile without it?

mulderr: It does. Making sure. I'll just go ahead and remove it.

mulderr:
I think (?) all the comments have been addressed in this latest commit. Not very descriptive but you probably want to squash anyway at this point. Tested again on linux, no issues.

Tried to follow style guide more closely but that 80 character maximum is tough.

runner/src/StudioRunner.hs
mulderr: Eliminated all use of `shell`.

runner/src/StudioRunner.hs
iamrecursion: Go for it!

runner/src/StudioRunner.hs
```diff
@@ -308,99 +304,91 @@ testIfRunning :: MonadRun m => m Bool
...
 
 startServices :: MonadRun m => m ()
 startServices = case currentHost of
     Windows -> do
-        path <- windowsScriptsPath
-        Shelly.shelly $ Shelly.silently $ Shelly.chdir path $ do
-            let startPath = path </> Shelly.fromText "start.bat"
-            Shelly.cmd startPath
+        path <- toFilePath <$> windowsScriptsPath
+        runProcess_ $ setWorkingDir path $ proc "start.bat" []
```
mwu-tow: This will fail to find `start.bat`, as the working directory is set for the spawned process but the lookup is done in the context of the current process. We should pass the full path to the `proc`. Changing working directory is not necessary, scripts use relative paths to their file locations AFAIK.

Also, I feel a little anxious when I see passing anything other than pe/coff image as the program path. But likely we have no better solution at hand. (I'd rather call to `cmd` giving it the path… but then it is even more risky, as user might have removed `cmd` from `Path`, so… let's leave it as is for now, that's total nitpick anyway)

mulderr: Hopefully fixed.

mwu-tow:
LGTM, well done and thank you for the contribution!

mulderr:
Thank you both for really helpful reviews as well!

iamrecursion:
And thank _you_ for your excellent contribution!

PR #1379, opened 2019-05-17
Fix backendLdLibraryPath

mulderr:
### Pull Request Description
Currently `backendLdLibraryPath` depends on CWD. It should instead reuse `absAppDir`.

If you build luna studio and start it from the `dist` directory then everything works fine. If however you start it from any other directory:
- some backend services fail to start as they cannot find zeromq libs: `exited: luna-double-representation (exit status 127; not expected)`
- atom starts but projects do not fully load because the backend isn't properly initialized (super confusing if the user doesn't notice errors in the console)

This PR fixes those issues.

### Important Notes
Nothing to report.

### Checklist

- [x] The code conforms to the [Haskell Style Guide](https://github.com/luna/luna/blob/master/doc/haskell-style-guide.md) and/or the [TypeScript Style Guide](https://github.com/luna/luna-studio/blob/master/doc/typescript-style-guide.md).

### DIFF
--- a/runner/src/StudioRunner.hs
+++ b/runner/src/StudioRunner.hs
@@ -130,6 +130,7 @@ getRunnerDir = liftIO $ do
     x <- getExecutablePath
     parent <$> parseAbsFile x
 
+-- | Path to application package root
 absAppDir :: MonadIO m => m (Path Abs Dir)
 absAppDir = do
     runnerDir <- getRunnerDir
@@ -143,6 +144,7 @@ absAppDir = do
                  else parent . parent . parent -- drop bin/public/luna-studio/
     pure $ stepUp runnerDir
 
+-- | Path to application directory inside user home directory
 absHomeDir ::  MonadRun m => m (Path Abs Dir)
 absHomeDir = do
     runnerCfg <- get @RunnerConfig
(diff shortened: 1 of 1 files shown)


runner/src/StudioRunner.hs
```diff
@@ -143,6 +144,7 @@ absAppDir = do
                  else parent . parent . parent -- drop bin/public/luna-studio/
     pure $ stepUp runnerDir
 
+-- | Absolute path to application resource directory inside user home directory
```
mwu-tow: I'd avoid "application resource directory" phrase, as we have `_resourcesFolder` that means something else. I'd go with something more like "application data directory" or just "application directory" (well, all directories contain data :P ).

runner/src/StudioRunner.hs
```diff
@@ -130,6 +130,7 @@ getRunnerDir = liftIO $ do
     x <- getExecutablePath
     parent <$> parseAbsFile x
 
+-- | Absolute path to application installation directory
```
mwu-tow: I wouldn't call this necessarily installation directory — IIRC on Linux it resolves to mounted appimage fs, not the place where appimage was installed.
I'd go more with something like "application package root".

Also, total nitpick, but: we don't need to write in comment that this is absolute path, if we have this explicitly coded both in the type signature and the function name.

runner/src/StudioRunner.hs
```diff
@@ -209,7 +211,7 @@ userStudioAtomHome = do
     baseDir   <- buildDirectoryPath absHomeDir [configHomeFolder, appName] >>= \x -> fmap (x </>) (parseRelDir =<< version)
     pure $ baseDir </> (runnerCfg ^. studioHome)
 backendLdLibraryPath = do
-    ldLibPath <- PIO.getCurrentDir
+    ldLibPath <- absAppDir
```
mwu-tow: While we are at this, let's fix the naming as well. Obviously application directory is not a `ldLibPath` — the returned value from function is.

mulderr:
Any news on this? Always happy to make more changes.

